PR #367, opened 2014-09-23
Performance improvement and fixed warnings

unsign3d:
Hi, I fixed an error on some not configured environment with php.ini default and changed the way api controller do the calls, now instead of linear progression trough the option in the switch all the function are stored in one array (in php is not an array but a dictionary), so in theory adding new services will only increase in a logarithmic fashion instead of linear.

I'm so sorry for the mess in the commits but I'm not 100% able with git :(


flodolo:
Hi, still interested in finishing this? Otherwise it doesn't make much sense to keep the PR open.

flodolo:
Closing this. Feel free to open a new PR.

PR #407, opened 2014-12-12
Fix issue 360 notify user when we don't display all the results

SkySymbol:


pascalchevrel:
@SkySymbol do you intend to finish this patch or should we take it over?

SkySymbol:
Yes, sorry, i had some exams last week and i have trouble using the heredoc syntax

pascalchevrel:
ok :) you can ping me on IRC if you need help

PR #419, opened 2015-01-15
make mozilla.org use en-US as default locale

pascalchevrel:


flodolo:
Looks good, I can't find any other reference to en-GB around.

PR #421, opened 2015-01-22
3.6 release

pascalchevrel:
* Make News page use MVC
* Add helper functions in model for News page to make updates easier

### DIFF
--- a/app/inc/dispatcher.php
+++ b/app/inc/dispatcher.php
@@ -21,8 +21,8 @@ switch ($url['path']) {
         $page_descr = 'One source locale, get search results for two target locales';
         break;
     case 'news':
-        $view = 'changelog';
-        $page_title = 'Transvision News. Version Notes';
+        $controller = 'changelog';
+        $page_title = 'Transvision News and Release Notes';
         $page_descr = '';
         $css_include = ['changelog.css'];
         break;
--- a/app/models/changelog.php
+++ b/app/models/changelog.php
@@ -1,0 +1,97 @@
+<?php
+namespace Transvision;
+
+$releases = [
+    '1.0'   => '2012-07-27',
+    '1.1'   => '2012-08-01',
+    '1.2'   => '2012-08-10',
+    '1.3'   => '2012-08-17',
+    '1.4'   => '2012-09-04',
+    '1.5'   => '2012-10-02',
+    '1.6'   => '2012-10-18',
+    '1.7'   => '2012-10-24',
+    '1.8'   => '2013-01-04',
+    '1.9'   => '2013-01-11',
+    '2.0'   => '2013-01-18',
+    '2.1'   => '2013-01-30',
+    '2.2'   => '2013-02-28',
+    '2.3'   => '2013-03-22',
+    '2.4'   => '2013-04-10',
+    '2.5'   => '2013-04-18',
+    '2.6'   => '2013-06-14',
+    '2.7'   => '2013-07-05',
+    '2.8'   => '2013-08-09',
+    '2.9'   => '2013-10-26',
+    '3.0'   => '2013-12-18',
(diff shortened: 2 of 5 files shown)


app/inc/dispatcher.php
```diff
@@ -21,7 +21,7 @@
         $page_descr = 'One source locale, get search results for two target locales';
         break;
     case 'news':
-        $view = 'changelog';
+        $controller = 'changelog';
         $page_title = 'Transvision News. Version Notes';
```
flodolo: Never really thought about this title. Maybe `Transvision News and Release Notes`?

pascalchevrel: Sounds good,will change that in a minute

app/models/changelog.php
```diff
@@ -0,0 +1,98 @@
...
+        default:
+            $type = '';
+            $text = 'Other';
+            break;
+    }
+    return "<span class=\"release_tag {$type}\">{$text}</span> ";
+};
+
+// Helper to generate the list of patches for the version
+$github_link = function($release) use($releases) {
```
flodolo: We're not exactly consistent at the moment, but most of the code is using `function(vars) use (vars)` with a space after `use` (there are 2 cases without space). Same below.

app/models/changelog.php
```diff
@@ -0,0 +1,98 @@
...
+    $keys = array_keys($releases);
+    $previous = $keys[array_search($release, $keys) -1];
```
flodolo: nit: space before `1`

app/views/changelog.php
```diff
@@ -1,50 +1,80 @@
...
-?>
-<h2 class="release_number" id="v3.5.1"><a href="#v3.5.1">Version 3.5.1 <span class="release_date">2014-10-07</span></a></h2>
-
+<?=$release_title('3.6')?>
 <h3>End user visible changes</h3>
 <ul>
-    <li><?=$relnotes('bug')?>Fixed a bug that prevented new Gaia repositories – fetched from external web service – to appear in the list of available repositories when performing a search (flod)</li>
+    <li>
+        <?=$relnotes('new')?>
+        Added a new view exposing all <a href="/unchanged/">strings that were kept identical to en-US</a> (flod)
```
flodolo: were kept -> are kept

app/views/changelog.php
```diff
@@ -1,50 +1,80 @@
...
+        <?=$issue(413)?>
+        Replace Kint library by Symfony/var_dumper library for quick debugging, new <code>dump()</code> and <code>cli_dump()</code> helper functions (pascal)
+    </li>
+    <li>
+        <?=$relnotes('bug')?>
+        Archlinux compatibility fix: Force the use of Python 2 and not 3 (SkySymbol)
+    </li>
+    <li>
+        <?=$relnotes('change')?>
+        Bugzilla and  Cache classes were moved as external libraries for sharing with other l10n-drivers PHP tools (flod)
```
flodolo: not really used, but double space before Cache

app/models/changelog.php
pascalchevrel: Just checked the PSR recommandation and they advice adding spaces:
https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#6-closures
"Closures MUST be declared with a space after the function keyword, and a space before and after the use keyword."

flodolo:
I would add #390, maybe in the API section even if it affects also website queries. 
Example: "Queries for mozilla.org don't return English text when the string is untranslated". 

We lost a section for 3.5.1: is that wanted?

app/models/changelog.php
flodolo: Interesting. We're definitely missing a space after `function` in a ton of places

PR #424, opened 2015-01-26
Make Transvision usable without JavaScript enabled

flodolo:
Also:
* move JavaScript to the bottom of the template.
* fix ‘search type’ select at small views.


pascalchevrel:
I am rerunning the job on Travis as it failed but is probably a problem on Travis side

flodolo:
I was looking at the log: is it common to display a warning about composer being out-of-date? Never noticed it before.

pascalchevrel:
Yes it's common, the warning happens when you have a version of composer older than a month, Travis has composer pre-installed but they don't update it regularly

Issue #427, opened 2015-02-02
Update composer.json to follow security updates for dependencies


pascalchevrel:
Using https://security.sensiolabs.org/check I see that one of our dependencies is not recent enough and has a security issue (monolog).

I recently watched a presentation by the Composer author and he advices to use the tilde operator almost systematically to make sure we have updated dependencies.

I am going to propose a composer.json patch to follow this advice.

PR #428, opened 2015-02-02
Issue #427: update the way we fetch dependencies in composer

pascalchevrel:
- use tilde operator instead of *
- by default, check out all feature and security updates (1.x.x) but not API breaks (1.x going to 2.x)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,14 +15,14 @@
     },
 
     "require": {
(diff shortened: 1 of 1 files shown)


flodolo:
I find the ~ slightly confusing compared to *, but tested locally and works as expected.

PR #430, opened 2015-02-04
Add link to API request in onestring view

flodolo:
Small promo to our APIs. Not sure if it makes sense to add it to some other views.


pascalchevrel:
I think that it is a good idea but that we could make it more generic by just appending ?json to the query url and the redirect will take place, that way, we could include that in any view as a sub-template and not just in the onestring view

flodolo:
As discussed on IRC, closing this one in favor of a more generic approach. We'll eventually reuse what's useful from this one.

Issue #431, opened 2015-02-04
Promote our API on search result pages that have an API equivalent 


pascalchevrel:
Implement Flod's idea :)

PR #432, opened 2015-02-04
Issue #431: promote API version of search pages

pascalchevrel:
* creation of a new Utils::redirectToAPI() function
* new views/templates/ folder, template.php is now there and named base.php
* Onestring and main search views updated to diplay the API promo


flodolo:
Works great :+1:  

I'm a bit torn on displaying the message on two lines: it works in onestring view, a bit less in searches.
But at the same I don't like two different links being separated by a single period. So, it's either 2 lines or reduce the second link, i.e.
```
<p class="api_link">
    <span>API</span>These results are also available as an <a href="<?=\Transvision\Utils::redirectToAPI()?>">API request</a>.
    Learn more about <a href="https://github.com/mozfr/transvision/wiki/JSON-API">Transvision API</a>.
</p>
```
I leave the final choice to you ;-)

pascalchevrel:
I am keeping the lines as separate, Maybe people think the two links are the same with they are next to each other

Issue #438, opened 2015-02-18
[Suggestion] Add Number of results found in "Variables Overview" view


anefzaoui:
e.g. http://transvision.mozfr.org/variables/?locale=ar&repo=gaia
I'd like to see how many results are shown so after I fix the mistaken variables and refresh I can see the new number and know there has been a difference.
This is useful especially when the number of errors is big.

SkySymbol:
Can I take this issue ?

pascalchevrel:
all yours :)

PR #441, opened 2015-03-09
Fix variable overview issue #438

SkySymbol:


app/views/checkvariables.php
```diff
@@ -35,11 +35,12 @@
 $target = array_map(['Transvision\AnalyseStrings', 'cleanUpEntities'], $target);
 
 $mismatch = AnalyseStrings::differences($source, $target, $repo);
-
+$mismatch_number = count($mismatch);
```
pascalchevrel: if you use it only once in the file, no need to create a variable

app/views/checkvariables.php
```diff
@@ -35,11 +35,12 @@
...
 $bugzilla_link = 'https://bugzilla.mozilla.org/enter_bug.cgi?format=__default__&component='
                . Bugzilla::getURLencodedBugzillaLocale($locale, 'products')
                . '&product=Mozilla%20Localizations&status_whiteboard=%5Btransvision-feedback%5D';
 
+echo "<h2>" . $mismatch_number . " results found</h2>";
```
pascalchevrel: 4 comments:
1/ echo is fine but we tend to use "print" in out codebase
2/ don't use double quotes if you are not going to extrapolate variables
3/ since you use the mismatch_number only once, you can save a variable and out the count directly in the concatenation:
```php
print '<h2>' . count($mismatch) . ' results found</h2>';
```
4/ nitpick, what happens if you only have one result ? Your string would have a syntax error (1 results found) ;)

PR #442, opened 2015-03-19
fix the variableoverview

SkySymbol:
add a counter in the variable overview


app/views/checkvariables.php
```diff
@@ -40,6 +40,12 @@
                . Bugzilla::getURLencodedBugzillaLocale($locale, 'products')
                . '&product=Mozilla%20Localizations&status_whiteboard=%5Btransvision-feedback%5D';
 
+print '<h2>' . count($mismatch)
+    . (count($mismatch) == 1
+    ? ' result'
+    : ' results'
+    . ' found</h2>');
```
pascalchevrel: ending parentesis seems wrong there, should be after 'results' as the objective of using parentesis here is to make the ternary test prioritary over concatenation.

Issue #337, opened 2014-08-23
In check variables view, ignore spaces after {{ and before }}


TheoChevalier:
When localizers add or remove spaces around variable name, we report the string as potentially wrong, while it's correct.

We should not report the following cases:
en-US: {{ n }} selected
locale: {{n}} zaznaczone

en-US: {{n}} selected
locale: {{ n }} zaznaczone

PR #446, opened 2015-04-04
 Improve check variables controls, switch to MVC

flodolo:
* We need l10njs checks also in browser (for PDFViewer and Loop)
* l10njs: _ is a valid character for a variable name
* l10njs: ignore white spaces ({{n}} == {{ n }})
* printf: discovered that %s is not the same as %S: https://bugzilla.mozilla.org/show_bug.cgi?id=1133554
* Use regular expression to check translations. Example: en-US has "$name", translation has "$nameDA", error is not reported with strpos.

Switched view to MVC and added component filter to view.


flodolo:
This fixes #337, which is not really an easy pick.

flodolo:
Using this PR to switch to MCV too, so WIP for now (we need a tag for that).

flodolo:
This is really weird. I pushed several times to my branch, but it's not reflected in this PR :8ball:

flodolo:
Got a reply from Github support (that was fast, especially on Easter morning!). It's an issue on their side, they're working on it.

flodolo:
OK, PR updated, ready to test and review.

pascalchevrel:
Works well, LGTM

Issue #452, opened 2015-04-19
Changes to base code to use https for mozfr references


pascalchevrel:
- Use https for html content (currently favicon is loading from http) to avoid mixed content warnings
- Always use https when referencing Transvision URLs (In Bugzilla references for example)
- Choose between http and https depending on the server for API redirects, needed for local development as we use http in local and https on remote

Patch ready

PR #453, opened 2015-04-19
Issue #452: make Transvision HTTPS friendly

pascalchevrel:


flodolo:
There's one link to fix in README.md, and others in changelog.php (we should switch to `//`).

app/controllers/mainsearch.php
```diff
@@ -34,9 +34,12 @@
     $regex['perfect'] = isset($_GET['perfect_match']) ? 'perfect_match=1' : '';
     $regex = array_filter($regex);
     $regex = count($regex) > 0 ? '?' . implode('&', $regex) : '';
+    $scheme = isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off'
```
flodolo: I think it should be safe to move this to a shared file (e.g. ```/inc/router.php```). Maybe ```transvision_scheme``` as a name?

PR #460, opened 2015-05-05
Check variables (l10n.js): ignore non-breaking whitespaces

flodolo:
\s without \u doesn’t catch non-breaking spaces
http://stackoverflow.com/questions/3230623/filter-all-types-of-whitespac
e-in-php

Also fixed the regexp to return the variable name as group.


flodolo:
Example of failing check (3rd result)
https://transvision.mozfr.org/variables/?locale=bn-IN&repo=gaia

PR #462, opened 2015-05-09
Prevent HTML hightlight code from being escaped on entity results vie…

### LINKED ISSUE #461: Results are displaying HTML added by Transvision when searching for entities
When we are searching for entity name (search_type=entities), the HTML code used to display a fake non-breaking space (<span class="highlight-gray"> </span>) is displayed in the page, like in this one:

https://transvision.mozfr.org/?sourcelocale=en-US&locale=fr&repo=central&search_type=entities&recherche=suite/chrome/common/notification.properties:vulnerableNoUpdatePluginWarning

The same string looks fine while searching the string itself: https://transvision.mozfr.org/?recherche=Vulnerable+Plugin!&repo=central&sourcelocale=en-US&locale=fr&search_type=strings

I think it can be reproduced with any string containing thin or regular nbsp.
"search_type=strings_entities" does not seem affected

TheoChevalier:
…w - fixes #461

### DIFF
--- a/app/views/results_entities.php
+++ b/app/views/results_entities.php
@@ -23,12 +23,18 @@ foreach ($entities as $entity) {
         $path_locale2 = VersionControl::hgPath($locale, $check['repo'], $entity);
     }
 
-    $target_string2 = '';
+    // Escape strings for HTML Display
+    $bz_target_string = $target_string = isset($tmx_target[$entity])
+                                            ? Utils::secureText($tmx_target[$entity])
+                                            : '';
+    $bz_target_string2 = $target_string2 = isset($tmx_target2[$entity])
+                                            ? Utils::secureText($tmx_target2[$entity])
+                                            : '';
+    $source_string = Utils::secureText($tmx_source[$entity]);
+
     if ($url['path'] == '3locales') {
-        if (isset($tmx_target2[$entity])) {
-            // nbsp highlight
-            $target_string2 = str_replace(' ', '<span class="highlight-gray"> </span>', $tmx_target2[$entity]);
-        }
+        // nbsp highlight only after strings have been escaped
+        $target_string2 = str_replace(' ', '<span class="highlight-gray"> </span>', $target_string2);
 
         if ($check['repo'] == 'mozilla_org') {
             $path_locale3 = VersionControl::svnPath($locale2, $check['repo'], $entity);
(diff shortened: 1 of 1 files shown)


TheoChevalier:
Crossing my fingers that this is not causing more regressions than that it's fixing, this code area looks… sensitive :)

TheoChevalier:
I knew I'd forgot something… When filing a bug from entity view, the string contains HTML code.

app/views/results_entities.php
```diff
@@ -23,12 +23,18 @@
         $path_locale2 = VersionControl::hgPath($locale, $check['repo'], $entity);
     }
 
-    $target_string2 = '';
+    // Escape strings for HTML Display
```
flodolo: display (lowercase)

app/views/results_entities.php
```diff
@@ -23,12 +23,18 @@
...
+    $bz_target_string = $target_string = isset($tmx_target[$entity])
+                                            ? Utils::secureText($tmx_target[$entity])
+                                            : '';
+    $bz_target_string2 = $target_string2 = isset($tmx_target2[$entity])
```
flodolo: You should put this definition in the following if cycle, we don't need it for the normal view, only for 3locales.

app/views/results_entities.php
```diff
@@ -23,12 +23,18 @@
...
+                                            ? Utils::secureText($tmx_target2[$entity])
+                                            : '';
+    $source_string = Utils::secureText($tmx_source[$entity]);
+
     if ($url['path'] == '3locales') {
-        if (isset($tmx_target2[$entity])) {
-            // nbsp highlight
-            $target_string2 = str_replace(' ', '<span class="highlight-gray"> </span>', $tmx_target2[$entity]);
-        }
+        // nbsp highlight only after strings have been escaped
```
flodolo: ```// Highlight non-breaking spaces``

app/views/results_entities.php
```diff
@@ -37,17 +43,8 @@
...
-    } else {
-        $target_string = '';
-    }
-
-    // Escape strings for HTML Display
-    $source_string  = Utils::secureText($tmx_source[$entity]);
-    $target_string  = Utils::secureText($target_string);
-    $target_string2 = Utils::secureText($target_string2);
+    // nbsp highlight only after strings have been escaped
+    $target_string = str_replace(' ', '<span class="highlight-gray"> </span>', $target_string);
```
flodolo: Same note about the comment, I would move this one to line 30 where you define `$bz_target_string = $target_string`

TheoChevalier:
Thanks :) Should be fixed

PR #463, opened 2015-05-09
Allow developers to toggle PERF_CHECK from config.ini

TheoChevalier:
Maybe I'm missing something, but currently to enable PERF_CHECK I edit a file tracked by Git (constants.php). And I commit this change by mistake every. single. time. :)

This PR allows us to control PERF_CHECK from an untracked file (config.ini), this way we can't commit PERF_CHECK = true.

If you prefer not to add more prefs to config.ini-dist, we could just make the change in constants.php so that we can still put the pref in our local config.ini ourselves.

Let me know what you guys think

### DIFF
--- a/app/config/config.ini-dev
+++ b/app/config/config.ini-dev
@@ -26,3 +26,6 @@ libraries=/temp
 
 ; Flag to know if we are working in production more or development mode
 dev=true
+
+; Flag to display page time generation and memory usage in console logs (development mode only)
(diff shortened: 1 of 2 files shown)


flodolo:
Let’s see what Pascal thinks, but it makes a lot of sense to me. We shouldn't have customizable settings in files tracked by Git.

pascalchevrel:
LGTM

PR #468, opened 2015-05-13
Fixing Select button on TMX download view

TheoChevalier:


flodolo:
Another fallout of JavaScript reorganization :-1: 

LGTM

flodolo:
Also checked the rest of the files, we don't have any other `<script>` around anymore.

Issue #471, opened 2015-06-03
Create a "switch source<->target" feature


goofy-bz:
It would be cool to switch from source to target lang and back with a simple tiny bidirectional arrow icon in between the two dropdown menus.

pascalchevrel:
done

PR #478, opened 2015-06-12
Issue #471: Create a "switch source<->target" feature

pascalchevrel:
+ also improves search form elements in mobile mode (<460px)


app/views/search_form.php
```diff
@@ -62,6 +62,9 @@ class="mainsearch_default_checkbox"
                          /> <span>Default</span>
                      </label>
                 </fieldset>
+                <fieldset id="locale_switch">
+                    <img src="/img/ic_swap_horiz_black_18dp.png" alt="Locale switcher" title="Locale switcher">
```
flodolo: Rename the PNG file?

web/js/main_search.js
```diff
@@ -77,4 +77,12 @@ $(document).ready(function() {
                           ';expires=' + expire.toGMTString();
         $(this).val(cookie_value);
     });
+
+    // Switch source and target locales in select boxes
+    $('#locale_switch').on('click', function() {
+      source_locale = $('select[id=source_locale]').val()
+      target_locale = $('select[id=target_locale]').val()
```
flodolo: Using `#source_locale` and Using `#target_locale` should be enough.

web/style/transvision.css
```diff
@@ -498,6 +498,14 @@ fieldset {
     padding-top: 0.2em;
 }
 
+#main_search fieldset#locale_switch {
+    margin:0 -6px;
+    padding:2.5em 0;
+    min-height:18px;
+    vertical-align: bottom;
```
flodolo: Space after `:` (first 3 lines)

app/views/search_form.php
flodolo: I would also use "Switch source and target locales" as title.

web/style/transvision.css
```diff
@@ -869,6 +877,27 @@ fieldset {
...
+    #main_search fieldset select,
+    #main_search fieldset select#search_type {
+        display: inline-block;
+        width: 13em;
+    }
+
+    #main_search fieldset#locale_switch {
+        margin: -1.2em 0;
+        padding: 0 0 0 14.5em;
+        height:18x;
```
flodolo: Space after `:`

web/style/transvision.css
```diff
@@ -869,6 +877,27 @@ fieldset {
...
+        cursor: pointer;
+    }
+    #main_search fieldset#locale_switch img {
```
flodolo: Empty line before the rule

flodolo:
Besides minor code nits, I wonder if we should create an element with fixed width/height and set the image as background. That would let use a different image < 460px (vertical arrows)

schermata 2015-06-12 alle 15 55 17

pascalchevrel:
PR updated to address review suggestions + use different backgroud images for swap icons depending on viewport width

Issue #482, opened 2015-06-17
Add a clear search button to the search bar


pascalchevrel:
This is a request by Sander Lepik, since he uses the middle-click button to copy/paste on Linux, he would like to be able to clear our search bar with the mouse too. I think it's a good idea.

PR #483, opened 2015-06-17
Issue #482: add a 'clear search' button on the input field

pascalchevrel:


web/js/main_search.js
```diff
@@ -83,9 +83,36 @@ $(document).ready(function() {
...
+            $(this).data('lastval', this.value);
+            $('#clear').show();
+        };
+
+        // If we delete the text with backspace, hide the button
+        if (! this.value) {
+            $('#clear').hide();
+        };
+    });
+
```
flodolo: Unless I'm missing something, you should be able to replace line 92-110 with
```
    // Hide the clear button if the search field is empty
    $('#recherche').on('input', function() {
        if (!this.value) {
            $('#clear').hide();
        } else {
            $('#clear').show();
        }
    });
```
And adding ```display: none``` in the CSS for `#clear` (when loading the page the field is empty).

Side note: we're not using a space after ! in JS, also don't need the ; after an `if`.

app/views/search_form.php
```diff
@@ -10,14 +10,17 @@
...
+                <div id="search_field_container">
+                    <input type="text"
+                           name="recherche"
+                           id="recherche"
+                           value="<?=$initial_search?>"
+                           placeholder="Type your search here"
+                           title="Type your search here"
+                           size="30"
+                    />
+                    <span id="clear"></span>
```
flodolo: Not the best in terms of accessibility, but I would add a title like "Clear the search field" to the span.
Also, `clear_search` as ID? `clear` seems a bit too generic.

web/style/transvision.css
```diff
@@ -507,6 +508,20 @@ fieldset {
...
+    position: relative;
+    display: inline;
+}
+
+#clear {
+    position: absolute;
+    top: 0;
+    right: 0;
+    width: 40px;
+    height: 22px;
```
flodolo: When resized to smaller viewports the X is visibly off-center vertically. We can use 20px for all sizes, or add rules to viewports and keep 22px for the big one.
schermata 2015-06-17 alle 14 36 03

app/views/search_form.php
```diff
@@ -10,14 +10,17 @@
...
+                <div id="search_field_container">
+                    <input type="text"
+                           name="recherche"
+                           id="recherche"
+                           value="<?=$initial_search?>"
+                           placeholder="Type your search here"
+                           title="Type your search here"
+                           size="30"
+                    />
+                    <span id="clear_search" alt="Clear the search field"></span>
```
flodolo: I meant `title` attribute, not `alt`.

Issue #486, opened 2015-06-17
Bust  cached javascript ressources with each new release


Archaeopteryx:
Enabling the checkbox to set the value of 'Search In' as default doesn't get used as default value the next time transvision gets loaded. Test with Firefox 38.0.5

flodolo:
It's a cache issue (goes away with a forced reload). Not sure if we want to be less aggressive with cache
https://github.com/mozfr/transvision/blob/master/web/.htaccess#L15-L22

pascalchevrel:
We can add a query parameter with the version number, this way the http cache will be busted with each release. Looking at the code, that's what we are already doing with css, reopening and assigning to myself.

pascalchevrel:
fixed

PR #487, opened 2015-06-18
Issue #486: bust js and css caches with new release

pascalchevrel:


app/views/templates/base.php
```diff
@@ -112,30 +122,29 @@
 
   <div id="footer">
     <p>Transvision is a tool provided by the French Mozilla community, <a href="https://www.mozfr.org" title="Home of MozFR, the French Mozilla Community" hreflang="fr">MozFR</a>.</p>
-    <?php echo $last_update; ?>
+    <?= $last_update ?>
```
flodolo: Do we want spaces or not? We're using a bit of both (mostly not I think)

app/views/templates/base.php
```diff
@@ -65,16 +78,13 @@
...
 } ?><?=$title_productname?></title>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <link rel="stylesheet" href="/style/transvision.css?<?php echo VERSION; ?>" type="text/css" media="all" />
-    <?php
-    foreach ($css_include as $css_file) {
-        echo "<link rel=\"stylesheet\" href=\"/style/{$css_file}?" . VERSION . "\" type=\"text/css\" media=\"all\" />\n";
-    }
-    ?>
+<?php foreach ($css_include as $css_file):?>
```
flodolo: We should stick to the syntax with parenthesis for consistency `foreach () { ... }`

flodolo:
php-cs to fix, but the rest looks good.

app/views/templates/base.php
pascalchevrel: I don't have a strong preference, I am thinking that spaces makes reading the variable name maybe more easily maybe? I let you choose on this one and I'll update the file for whatever choice we do.

app/views/templates/base.php
pascalchevrel: I don't agree, php templating syntax is easier to read and less verbose since we don't have to output strings or escape quotes when we output html and this file *is* a template.

flodolo: To be honest I'm not a huge fan of mixing so much HTML/PHP, or of `<?=$var?>`, but I guess I can live with it.

app/views/templates/base.php
flodolo: OK for spaces then.

app/views/templates/base.php
pascalchevrel: I am experimenting with Twig these days (it is Jinja for PHP, originally by the author of Jinja), eventually we could switch to it but it will require rewriting all views to use that.

PR #492, opened 2015-06-30
Restructure TMX folders

flodolo:
Move from /TMX/repo/locale to /TMX/locale

setup.sh
* Check existence of /TMX/locale
* Switch URLs to https

tmxmaker.py
* Use meaningful variable names
* Use argparse instead of OptionParser (deprecated since 2.7)
* Store cache files in TMX/locale, don't create duplicated en-US cache files

mozorg.php
* Move script from /app/inc/mozorg.php to /app/scripts/tmx_mozillaorg
* Generate cache file in /TMX/locale/cache_locale_mozilla_org.php

Utils::getRepoStrings
* Update cache file path and tests

Dotlang::getStrings
* Manage reference locale differently (store original string instead of empty)


flodolo:
For testing I renamed the existing TMX folder.

Original TMX: 1.8 GB (1559 files)
New TMX: 903 MB (785 files)

app/scripts/glossaire.sh
```diff
@@ -278,7 +278,7 @@ function updateFromSVN() {
     then
         echogreen "Extract strings on svn"
         cd $install
-        nice -20 php app/inc/mozorg.php
+        nice -20 php app/scripts/tmx_mozillaorg
```
pascalchevrel: since you added a php shebang to tmx_mozillaorg you probably don't need to explicitely call the php interpreter if you make tmx_mozillaorg executable

flodolo:
Fixed php script calls, left in a separate commit.

pascalchevrel:
I did an install from scratch with your patch and it works fine, so it's ok for me.

Issue #493, opened 2015-07-04
Add Piwik stats


pascalchevrel:
We have set up Piwik on the mozfr server, we can activate it for the Transvision subdomain.

That means that we need to add some js in the base template to send hits to Piwik.

For pages without a template (our JSON API calls), I need to investigate if/how those can be counted

We also need to make it an optional setting because we don't want local installs, or non-official installs (on our VPSes or other people's) to send data to our Piwik instance. Similarily, we probably want other installations to send data to a separate Piwik instance or a different profile (for example a separate stats profile for our beta site).

pascalchevrel:
That also means that we will have to apply the patch on the production server before a release so as to make sure that it works and sends data to Piwik correctly

flodolo:
Add a `piwikurl` to config.ini and inject JS only if that's available?

pascalchevrel:
That's the patch I am doing now :)

pascalchevrel:
I couldn't find any way to track non-HTML pages

PR #494, opened 2015-07-04
Issue #493: Piwik Template integration

pascalchevrel:
- add piwik_url and piwik_id variables to config.ini
- integrate js only in production mode, if the settings exist and are valid

### DIFF
--- a/app/inc/constants.php
+++ b/app/inc/constants.php
@@ -18,7 +18,7 @@ define('MODELS',        APP_ROOT . 'models/');
 define('CONTROLLERS',   APP_ROOT . 'controllers/');
 define('CACHE_ENABLED', isset($_GET['nocache']) ? false : true);
 define('CACHE_PATH',    INSTALL_ROOT . 'cache/');
-define('APP_SCHEME', isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off' ? 'https://' : 'http://');
+define('APP_SCHEME',    isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off' ? 'https://' : 'http://');
 
 if (file_exists(CACHE_PATH . 'lastdataupdate.txt')) {
     define('CACHE_TIME',  time() - filemtime(CACHE_PATH . 'lastdataupdate.txt'));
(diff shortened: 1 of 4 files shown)


app/inc/constants.php
```diff
@@ -28,7 +28,8 @@
 }
 
 // Special modes for the app
-define('DEBUG', (strstr(VERSION, 'dev') || isset($_GET['debug'])) ? true : false);
+define('DEBUG',     strstr(VERSION, 'dev') || isset($_GET['debug']));
+define('LOCAL_DEV', isset($server_config['dev']) && $server_config['dev']);
```
flodolo: I'm lost, where should `$server_config['dev']` come from?

pascalchevrel: https://github.com/mozfr/transvision/blob/master/app/config/config.ini-dev#L27

app/views/templates/piwik.php
```diff
@@ -0,0 +1,25 @@
...
+  _paq.push(['enableLinkTracking']);
+  (function() {
+    var u="<?=$server_config['piwik_url']?>";
+    _paq.push(['setTrackerUrl', u+'piwik.php']);
+    _paq.push(['setSiteId', <?=$server_config['piwik_id']?>]);
+    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
+  })();
+</script>
+<noscript><p><img src="<?=$server_config['piwik_url']?>piwik.php?idsite=<?=$server_config['piwik_id']?>" style="border:0;" alt="" /></p></noscript>
```
flodolo: Should you have a '/` before `piwik.php`? Anyway worth a note in the comment in config.ini-dist, if the URL has to have a trailing slash or not.

pascalchevrel: it needs a trailing slash, will add a note

app/inc/constants.php
flodolo: Wow, never noticed it. Are we already using it somewhere? Can't find anything in the code.

Can you add `dev=false`to config.ini-dist? I also wonder if we should just use one system to determine if it's dev or not (I always considered the version to be the trigger, but I like a specific parameter a lot more).

flodolo:
Besides my confusion about `dev` in ```config.ini``` (which can be fixed/discussed somewhere else), this should be good to merge.

app/inc/constants.php
pascalchevrel: we use it in the dev-setup script to install or not a snapshot:
https://github.com/mozfr/transvision/blob/master/app/scripts/dev-setup.sh#L79

Yes, I can add dev=false to -dist

There is a difference between a dev setup and a dev version (in that context it means beta), maybe we should use beta instead but I would prefer that to be part of a separate issue.

pascalchevrel: actually, dev=false is already in config.ini-dist:
https://github.com/mozfr/transvision/blob/master/app/config/config.ini-dist#L32

Issue #497, opened 2015-07-07
Search fail for entities+perfect match


flodolo:
https://transvision.mozfr.org/?recherche=sourceNameIE&repo=aurora&sourcelocale=en-US&locale=de&search_type=entities&perfect_match=perfect_match

flodolo:
So, the problem is that we consider `browser/chrome/browser/migration/migration.properties:sourceNameIE` as the perfect match, not just `sourceNameIE`, and that's confusing.

flodolo:
Looking into this. I think I've figured out the search results part, not yet the API side.

PR #498, opened 2015-07-07
Search only in entity names, not the full entity

flodolo:


flodolo:
This should fix issue #497

flodolo:
Still WIP (after looking at the diff I realized there's still a lot of duplicated code).

flodolo:
Closing this for now because there's a log more work to do, and don't want travis to run at every push.

PR #499, opened 2015-07-08
Improve search for entities

### LINKED ISSUE #497: Search fail for entities+perfect match

flodolo:
* Search also in entity names if search on full entity fails (useful for "perfect match")
* Don't display the table header if there are no results

Impact in terms of performance is small, since we search through entity names only when the first search fails.

Fixes #497

### DIFF
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -162,7 +162,7 @@ class ShowResults
     {
         $replacements = [
             ' '  => '<span class="highlight-gray"> </span>',
-            '…' => '<span class="highlight-gray">…</span>',
+            '…'  => '<span class="highlight-gray">…</span>',
         ];
 
         switch ($locale) {
@@ -217,9 +217,9 @@ class ShowResults
 
             // Don't highlight search matches in entities when searching strings
             if ($search_options['search_type'] == 'strings') {
-                $result_entity = ShowResults::formatEntity($key);
+                $result_entity = Self::formatEntity($key);
             } else {
-                $result_entity = ShowResults::formatEntity($key, $recherche[0]);
+                $result_entity = Self::formatEntity($key, $recherche[0]);
             }
 
             $component = explode('/', $key)[0];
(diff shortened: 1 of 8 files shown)


flodolo:
@pascalchevrel Ping?

pascalchevrel:
forgot about this one, sorry. Looking at it now

app/classes/Transvision/ShowResults.php
```diff
@@ -403,4 +403,35 @@ public static function resultsTable($search_results, $recherche, $locale1, $loca
...
+        $entities = preg_grep($regex, array_keys($source_strings));
+
+        /* If there are no results, search also through the entity names.
+         * This is needed for "perfect match" when only the entity name is
+         * provided.
+         */
+        if (! $entities) {
+            $entity_names = [];
+            foreach ($source_strings as $entity => $translation) {
+                $entity_name = explode(":", $entity)[1];
```
pascalchevrel: single quotes

app/classes/Transvision/ShowResults.php
```diff
@@ -403,4 +403,35 @@ public static function resultsTable($search_results, $recherche, $locale1, $loca
...
+    public static function searchEntities($source_strings, $regex)
+    {
+        // Search through the full entity ID
+        $entities = preg_grep($regex, array_keys($source_strings));
+
+        /* If there are no results, search also through the entity names.
+         * This is needed for "perfect match" when only the entity name is
+         * provided.
+         */
+        if (! $entities) {
```
pascalchevrel: although it works because php is a dynamically typed language and an empty array is evaluated as false, I think I would prefer that we explicitely check for emptyness with empty()

app/classes/Transvision/ShowResults.php
```diff
@@ -403,4 +403,35 @@ public static function resultsTable($search_results, $recherche, $locale1, $loca
...
+        $entities = preg_grep($regex, array_keys($source_strings));
+
+        /* If there are no results, search also through the entity names.
+         * This is needed for "perfect match" when only the entity name is
+         * provided.
+         */
+        if (empty($entities)) {
+            $entity_names = [];
+            foreach ($source_strings as $entity => $translation) {
+                $entity_name = explode(':', $entity)[1];
```
pascalchevrel: you can skip the temp $entity_name variable here and just assign explode in the next line, you don't reuse this var

flodolo:
Weird, I have php-cs failure on the tests, need to test locally.

flodolo:
Oh boy, I've updated mozilla-l10n-query to add Gaia 2.2, need to fix that test too (php-cs is fixed now)

pascalchevrel:
the rest looks good to me

flodolo:
Thanks. Fixed the API functional test, waiting for Travis to be finally green before merging.

Issue #502, opened 2015-09-07
assets folder not created on a fresh install


pascalchevrel:
As discussed with @SkySymbol during the first install of Transvision, jquery is not installed because composer is set up to install the lib in web/assets which doesn't exist.

We agreed that a pre-install-cmd in composer.json is a solution we prefer over having a blank folder with a dummy file on Github https://getcomposer.org/doc/articles/scripts.md#command-events

Assigning to Vincent.

PR #503, opened 2015-09-07
add robloach/component-installer as a dependency #502

SkySymbol:
adding robloach/component-installer as a dependency to install jquery in the right directory


pascalchevrel:
Can you put it at the bottom of the require list please? We have the required dependencies listed alphabetically :)

PR #505, opened 2015-09-29
Update experimental status of views, fix Utils::getRepoStrings

flodolo:
Remove experimental status to old (tested) views.

Fix Utils::getRepoStrings in case of missing TMX: if the file is missing, method currently returns false ($tmx is always set, false if the Cache element doesn’t exist).

### DIFF
--- a/app/classes/Transvision/Utils.php
+++ b/app/classes/Transvision/Utils.php
@@ -239,7 +239,7 @@ class Utils
             }
         }
 
-        return isset($tmx) ? $tmx : [];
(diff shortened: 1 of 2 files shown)


flodolo:
Noticed the issue in `/stats`, log was full of error since I don't have TMX for gaia 2.2 for all locales.

One more change that I left out: should we remove the "Translate String" item in the top menu? It doesn't seem helpful, since you can't move out of that page, e.g. searching for a different string.

pascalchevrel:
I'll remove the Translate String menu item in a separate commit

PR #508, opened 2015-10-05
Issue 507:  Update Transvision for mozilla.org migration to Github

pascalchevrel:
- setup.sh, glossaire.sh and tmx_mozillaorg now deal with github
- New VersionControl::gitPath() method generating link to file on Github
- Some refactoring with a private VersionControl::cleanFilePath() method to put code in common between svnPath() and gitpath()
- Updated tests
- in constants.php, our paths are now using realpath() + '/', this way the app doesn't break if we forget a trailing slash in config.ini

Notes:
- I am keeping svnPath() because of Firefox for iOS
- I am not adding a test for cleanFilePath() because it is a private method and it's behaviour is tested indirectly by svnPath() and gitPath()

### DIFF
--- a/app/classes/Transvision/VersionControl.php
+++ b/app/classes/Transvision/VersionControl.php
@@ -20,7 +20,7 @@ class VersionControl
     {
         $vcs = [
             'hg'  => [],
-            'svn' => ['mozilla_org'],
+            'git' => ['mozilla_org'],
         ];
         $vcs['hg'] = array_merge(
             Project::getDesktopRepositories(),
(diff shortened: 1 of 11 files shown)


app/classes/Transvision/VersionControl.php
```diff
@@ -194,19 +194,51 @@ public static function hgPath($locale, $repo, $path)
...
-        array_shift($path);
-        $path          = implode('/', $path);
+        return 'http://viewvc.svn.mozilla.org/vc/'
+               . $file_path . '?view=markup';
+    }
 
-        if ($repo == 'mozilla_org') {
-            $url .= 'projects/mozilla.com/trunk/locales/' . $locale . '/' . $path;
+    /**
+     * Generate a path to the github repo for the file.
```
flodolo: Nit: GitHub

app/classes/Transvision/VersionControl.php
```diff
@@ -194,19 +194,51 @@ public static function hgPath($locale, $repo, $path)
...
-        return $url . '?view=markup';
+        return 'https://github.com/mozilla-l10n/'
+               . $repo . '/blob/master/'
+               . $locale . '/' . self::cleanFilePath($path);
+    }
+
+    /**
+     * Remove entity and project name from path
+     *
+     * @param  string $path A transvision file path
```
flodolo: Nit: Transvision.

Maybe extractFilePath is a clearer name for this method?

app/scripts/setup.sh
```diff
@@ -286,13 +286,13 @@ do
...
-echogreen "mozilla.org repo being checked out from subversion"
+# Check out GitHub repos
+echogreen "mozilla.org repo being checked out from GitHub"
 cd $mozilla_org
-if [ ! -d $mozilla_org/.svn ]
+if [ ! -d $mozilla_org/.git ]
 then
     echogreen "Checking out mozilla.org repo"
-    svn co https://svn.mozilla.org/projects/mozilla.com/trunk/locales/ .
+    git clone https://github.com/mozilla-l10n/www.mozilla.org .
```
flodolo: Should we do `git clone --depth 1` since we're not interested in the full history?

pascalchevrel: It's a one time cost and I think it could allow us to also query the history for a team (like author names), so i'd prefer not.

flodolo:
Tested locally, everything works as expected.

Issue #392, opened 2014-11-13
Translation memory API


mathjazz:
I'd like to query Translation memory API without specifying repository, meaning the entire Transvision database. Currently, separate API calls are needed to achieve that.

Might be useful addition for other API calls, too.

PR #516, opened 2015-10-14
Issue #392: Translation memory API, query all repositories

pascalchevrel:
- add a new alias for all repos called 'global'
- add Project::getRepositoriesLocale('ab-CD'), returns all the repositories available for a locale
- new public api/v1/repositories/<locale code>/ JSON API
- 'global' search added to translation memory API + general search (strings and entities) API
- Max results is now 500 for the API
- dupes are removed from the translation memory API
- update Project:getLocaleInContext() to have mappings for Spanishes
- add new unit and functional tests


app/classes/Transvision/Project.php
```diff
@@ -231,7 +255,12 @@ public static function getLocaleInContext($locale, $context)
         }
 
         // Firefox for iOS: no mapping
-        $locale_mappings['firefox_ios'] = [];
+        $locale_mappings['firefox_ios'] = [
```
flodolo: I'm just doing a preliminary swap of the code, but I don't understand this change.

We do have 3 [different translations for Spanish](https://l10n.mozilla-community.org/~flod/webstatus/?product=firefox-ios): es, es-CL, es-MX. With this we'll lose es-CL and es-MX.

pascalchevrel: my bad, I thought we had only es (that's what Inma told me...)

app/classes/Transvision/API.php
```diff
@@ -71,6 +71,8 @@ public function __construct($url)
         $this->extra_parameters = isset($url['query'])
             ? $this->getExtraParameters($url['query'])
             : [];
+
+        $this->valid_repositories =  array_merge(Project::getRepositories(), ['global']);
```
flodolo: Nit: double space before array_merge

app/classes/Transvision/API.php
```diff
@@ -302,11 +323,17 @@ private function verifyEnoughParameters($number)
...
      * @param  string  $repository Name of the repository
+     * @param  boolean $alias      Do we allow aliases for repository names,
+     *                             ex: 'global', to query all repositories. Default to False
      * @return boolean True if we support this repository, False if we don't
      */
-    private function verifyRepositoryExists($repository)
+    private function verifyRepositoryExists($repository, $alias = false)
     {
-        if (! in_array($repository, Project::getRepositories())) {
+        if (! in_array($repository, $this->valid_repositories)) {
```
flodolo: I think I lost your train of thoughts here. 

In the costructor you set `$this->valid_repositories` to all supported repositories plus `global`. So, `verifyRepositoryExists('global')` will return true, while it should return false, given that `global`is not an actual repository (e.g. `/api/v1/locales/global/` or `/api/v1/entity/global/?id=test`).

Also, you're not actually doing anything with the updated value of `$this->valid_repositories` after adding `global` if we accept an alias.

This method also seems to be the only user for `$this->valid_repositories`. Should we keep it as a class property?

I would expect something like this, without setting the value in the constructor
```
    private function verifyRepositoryExists($repository, $alias = false)
    {
        $this->valid_repositories = $alias
            ? array_merge(Project::getRepositories(), ['global'])
            : Project::getRepositories();
        if (! in_array($repository, $this->valid_repositories)) {
            $this->log("The repo queried ({$repository}) doesn't exist.");

            return false;
        }

        return true;
    }
```

app/classes/Transvision/Project.php
```diff
@@ -169,6 +170,29 @@ public static function getRepositoryLocales($repository)
...
+     *
+     * @param  string $locale Mozilla locale code
+     * @return array  A sorted list of repositories available for the locale
+     */
+    public static function getLocaleRepositories($locale)
+    {
+        $matches = [];
+        foreach (self::getRepositories() as $repository) {
+            if (in_array(
+                self::getLocaleInContext($locale, $repository),
```
flodolo: You want `$locale` here, not the contextualized locale code. `/api/v1/repositories/es/`should return the locales actually supported by `es`: `["firefox_ios","gaia","gaia_1_3","gaia_1_4","gaia_2_0","gaia_2_1","gaia_2_2"]`.

app/models/api/repositories_list.php
```diff
@@ -1,4 +1,10 @@
 <?php
 namespace Transvision;
 
+// We have a query for the repositories supported by a locale
+if (isset($request->parameters[2])) {
+    return $json = Project::getLocaleRepositories($request->parameters[2]);
+}
+
+// default to list all existing repositories
```
flodolo: Nit: Default

app/models/api/repository_search.php
```diff
@@ -12,25 +12,26 @@
...
+    : [$request->parameters[3]];
+
+$entities_merged       = [];
+$source_strings_merged = [];
+$target_strings_merged = [];
+
+// We loop through all repositories searched and merge results
+foreach ($repositories as $repository) {
+    $source_strings = Utils::getRepoStrings($request->parameters[4], $repository);
+    // $source_strings = array_merge($source_strings, Utils::getRepoStrings($request->parameters[4], $repository));
```
flodolo: Remove commented out code?

PR #528, opened 2015-11-10
Provide a way to go back to search from onestring view

flodolo:
I wonder if we could have issue by using an utf-8 character like 🔍

### DIFF
--- a/app/views/onestring.php
+++ b/app/views/onestring.php
@@ -23,13 +23,17 @@ include VIEWS . 'templates/api_promotion.php';
     <tr>
         <th>Locale</th>
         <th>Translation</th>
+        <th>&nbsp;</th>
     </tr>
     <?php
+    $reference_locale = Project::getReferenceLocale($repo);
     foreach ($translations as $locale => $translation) {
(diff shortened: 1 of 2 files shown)


app/views/onestring.php
```diff
@@ -23,13 +23,17 @@
     <tr>
         <th>Locale</th>
         <th>Translation</th>
+        <th>&nbsp;</th>
     </tr>
     <?php
+    $reference_locale = Project::getReferenceLocale($repo);
```
pascalchevrel: Maybe you don't need a temporary variable here since you only use it once later?

pascalchevrel:
All looks good, maybe a nit about a temporary variable but I can live with it :)

app/views/onestring.php
flodolo: It's used inside a foreach, which means we call the method once for every translation.

pascalchevrel: ok, I think it wouldn't be any problem with phpn 5.5 because it has a built-in opt-code cache but indeed we use 5.4 and we don't have one (yet)

PR #532, opened 2015-11-12
Try to fix intermittent failing tests

flodolo:


flodolo:
With the current code I believe we're never using the passed reference date.

If for example I set one the tests to
```
            [
                (new \DateTime('2013-01-12 12:00'))->modify('-2 seconds'),
                (new \DateTime('2013-01-12 12:00')),
                '2 seconds ago',
            ],
```
It fails with `'2 years ago' does not contain 2 seconds ago`.

pascalchevrel:
I think in both the Utils and tests/Utils class we should just import the DateTime class from the global scope:
```php
use \DateTime;
```

This way we can remove all the backslashes before the use of DateTime in the classes.

Also, if this is supposed to fix our Travis tests, we should maybe remove the tests explicitely excluded from Travis with : 
```php
        if (! getenv('TRAVIS')) {
```

Another thing I think is probably a small bug is this line in the tests:
```php
(new DateTime())->modify('-1 hours'),
```
I think 'hour' should be in singular.

flodolo:
Thanks, I should have addressed everything.

PR #538, opened 2015-11-14
Stop hard-coding version information in constants.php

flodolo:
* setup.sh: Move main thread code after function declarations
* setup.sh: Store version information in cache
* constants.php: Use cached information to determine BETA vs PROD version, update base.php template

The idea is to be able to update the production server without making an explicit release: simply tag the new changeset, create the release tag on GitHub and pull.

### DIFF
--- a/app/scripts/setup.sh
+++ b/app/scripts/setup.sh
@@ -23,38 +23,6 @@ function echogreen() {
     echo -e "$GREEN$*$NORMAL"
 }
 
-# Store current directory path to be able to call this script from anywhere
-DIR=$(dirname "$0")
-# Convert .ini file in bash variables
-eval $(cat $DIR/../config/config.ini | $DIR/ini_to_bash.py)
-
-# Generate sources (gaia versions, supported locales)
-echogreen "Generate list of locales and supported Gaia versions"
-$DIR/generate_sources $config
-
-# Check if we have sources
-echogreen "Checking if Transvision sources are available..."
-if ! $(ls $config/sources/*.txt &> /dev/null)
-then
-    echored "CRITICAL ERROR: no sources available, aborting."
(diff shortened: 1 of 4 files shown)


app/scripts/setup.sh
```diff
@@ -262,6 +230,54 @@ function initGaiaRepo () {
...
+for folder in "${folders[@]}"
+do
+    if [ ! -d $folder ]
+    then
+        echogreen "Creating folder: $folder"
+        mkdir -p "$folder"
+    fi
+done
+
+# Store version information in cache
```
flodolo: Real changes to setup.sh start from there

pascalchevrel:
There is always more than updating the version number for a release, that's actually the easiest part. There is also for example rerunning composer to update dependencies, so just that wouldn't make the stuff automatic. Also the cron job should be updated for a pull -all otherwise tags wouldn't be detected on a simple pull.

If you feel strongly about it it's ok, but I think it's not really fixing a real problem in shipping releases for Transvision, might save a few minutes a year, maybe :)

flodolo:
For me it's not much about gaining time, it's more about making the process more streamlined.

Right now, when we release we need to:
* Update the release number and remove dev
* Add a blurb to release notes
* Tag the release
* Update the server
* Update the release to a dev number. For example this cycle we haven't done that yet, and the [Beta site](https://transvision-beta.mozfr.org/) shows up as prod site.

And we need to do all this even for a dot release to fix a mistake (like 3.8.1).

We could simply create a script on the server, e.g. `release.sh`, that runs a full pull, update composer and runs setup. On the GitHub side we only need to tag the release, and update the release notes if we want too. I also wonder if we should update composer more frequently anyway.

Issue #530, opened 2015-11-12
Host our own fonts, don't get them from mozilla.org


pascalchevrel:
Out of convenience, we fefch fonts from mozilla.org but mozilla.org servers are in the US, slow for us localizers in EMEA and a lot of our initial page load time is getting them.

Let's host them ourselves on Transvision, potentially I might look at using Fira Sans instead of OpenSans since it seems mozilla is transitionning to that on the middle term.

PR #539, opened 2015-11-15
Issue #530: host OpenSans font ourselves

pascalchevrel:
* Don't provide fallbacks for old browsers, just use woff
* Cache them via .htaccess
* Put them in a local fonts folder

A quick check on my local instance gives good results on page load (/3)


web/style/transvision.css
```diff
@@ -1,33 +1,46 @@
 /* Fonts */
+/* Updated on www.mozilla.org 2015-06-17 */
 
 @font-face {
-    font-family: 'Open Sans';
+    font-family: 'Open Sans Light';
+    src: url('/fonts/OpenSans/OpenSans-Light-webfont.woff') format('woff');
```
flodolo: I'd remove quotes on the URL for consistency with the rest of the CSS.

web/.htaccess
```diff
@@ -13,11 +13,18 @@
...
     ExpiresByType image/jpeg              "access plus 1 month"
     ExpiresByType image/png               "access plus 1 month"
     ExpiresByType text/css                "access plus 1 year"
+    ExpiresByType application/font-woff   "access plus 1 year"
+    ExpiresByType application/javascript  "access plus 1 year"
+</IfModule>
+
+<IfModule mod_deflate.c>
+  <FilesMatch "\.(woff|svg)$" >
+    SetOutputFilter DEFLATE
```
flodolo: I confess I'm not familiar with this directive, but I wonder if it could be written like this (based on [this page](https://davidwalsh.name/html5-boilerplate-htaccess))
```
AddOutputFilterByType DEFLATE application/font-woff \
                              image/svg+xml
```

flodolo:
Besides the question about .htaccess, looks great to me :+1:

PR #545, opened 2015-12-07
Update Verif library to v0.3 with better looking output report

pascalchevrel:
See https://github.com/pascalchevrel/Verif/releases/tag/v0.3


pascalchevrel:
Here is how the new output looks like
verif03

PR #548, opened 2015-12-18
Fix nits in JS and CSS files

flodolo:
Found while looking in with Codacy


TheoChevalier:
Thanks, r+ :tropical_drink:

Issue #551, opened 2015-12-25
Add search suggestions as your type in main search from our own API


pascalchevrel:
I have a patch that allows using our own Transvision API to autocomplete suggestions as you type in the search bar.

PR #552, opened 2015-12-25
Issue #551: Search suggestion autocomplete

pascalchevrel:
* adds a new jquery libray that allows autocompletion
* base.php now allows defining files outside of the js folder (if the path starts with a /)

### DIFF
--- a/app/views/templates/base.php
+++ b/app/views/templates/base.php
@@ -7,7 +7,7 @@ $check['repo']  = isset($check['repo']) ? $check['repo'] : 'aurora';
 $source_locale  = isset($source_locale) ? $source_locale : 'en-US';
 $locale         = isset($locale) ? $locale : 'fr';
 $initial_search = isset($initial_search) ? $initial_search : 'Bookmarks';
-$base_js        = ['base.js'];
+$base_js        = ['/js/base.js'];
 $base_css       = ['transvision.css'];
 $cache_bust     = '?v=' . VERSION;
 
@@ -121,9 +121,8 @@ if (file_exists(CACHE_PATH . 'lastdataupdate.txt')) {
 
   <script src="/assets/jquery/jquery.min.js?v=<?= VERSION ?>"></script>
   <script src="/assets/clipboard.js/clipboard.js-built.js?v=<?= VERSION ?>"></script>
-
 <?php foreach ($javascript_include as $js_file):?>
-  <script src="/js/<?= $js_file . $cache_bust ?>"></script>
+  <script src="<?= $js_file . $cache_bust ?>"></script>
 <?php endforeach?>
 
   <script>
--- a/web/js/base.js
+++ b/web/js/base.js
@@ -13,8 +13,8 @@ $(document).ready(function() {
 
     // Associate code to link to hide/display top menu.
     $('#links-top-button').click(function(e) {
-      e.preventDefault();
-      hideMenuButton(400);
+        e.preventDefault();
+        hideMenuButton(400);
     });
 
     // Associate code to toggle search options on small screens.
--- a/web/style/transvision.css
+++ b/web/style/transvision.css
@@ -578,6 +578,37 @@ fieldset {
     width: 150px;
 }
 
+/* Autocomplete styles for suggestion popup */
+.autocomplete-suggestions {
+    border: 1px solid #999;
+    background-color: #fff;
+    overflow: auto;
+}
+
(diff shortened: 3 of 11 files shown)


app/views/templates/base.php
```diff
@@ -121,11 +121,12 @@
...
   <script src="/assets/clipboard.js/clipboard.js-built.js?v=<?= VERSION ?>"></script>
-
-<?php foreach ($javascript_include as $js_file):?>
-  <script src="/js/<?= $js_file . $cache_bust ?>"></script>
-<?php endforeach?>
-
+<?php
+foreach ($javascript_include as $js_file) {
+    $prefix = ($js_file[0] == '/') ? '' : '/js/';
+    print '  <script src="' . $prefix . $js_file . $cache_bust . '"></script>' . "\n";
```
flodolo: Should we just update the existing includes instead? There are only 5 or 6 of them

web/js/base.js
```diff
@@ -62,6 +62,30 @@ $(document).ready(function() {
 
     // Focus on the main search field.
     $('#recherche').focus();
+
+    // Autocomplete search suggestions in main search box using our own API
+    $('#recherche').autocomplete({
+        serviceUrl: function (query){
+            return 'https://transvision.mozfr.org/api/v1/tm/'
```
flodolo: We should query the local installation, not the remove prod server.

web/js/base.js
```diff
@@ -62,6 +62,30 @@ $(document).ready(function() {
...
+                + $('#repository').val() + '/'
+                + $('#source_locale').val() + '/'
+                + $('#target_locale').val() + '/'
+                + query + '/';
+        },
+        params: 'max_results=10',
+        minChars: 2,
+        transformResult: function(response) {
+            data = JSON.parse(response);
```
flodolo:
```
var data = JSON.parse(response);
```

web/js/base.js
```diff
@@ -62,6 +62,30 @@ $(document).ready(function() {
...
+            return {
+                suggestions: $.map(data, function(dataItem) {
+                    return { value: dataItem.source, data: dataItem.target };
```
flodolo: A find this style a bit clearer 
```JS
return { 
    value: dataItem.source, 
    data: dataItem.target 
};
```

web/js/base.js
```diff
@@ -62,6 +62,30 @@ $(document).ready(function() {
...
+                })
+            };
+        },
+        onSelect : function() {
```
flodolo: Nit: no space before colon

web/style/transvision.css
```diff
@@ -578,6 +578,37 @@ fieldset {
     width: 150px;
 }
 
+/* Autocomplete styles for suggestion popup */
+.autocomplete-suggestions {
+    border: 1px solid #999;
+    background: #FFF;
```
flodolo: HEX colors should be all lowercase for consistency, also `background-color` here is enough

web/style/transvision.css
```diff
@@ -578,6 +578,37 @@ fieldset {
...
+}
+
+.autocomplete-suggestion {
+    padding: 2px 5px;
+    white-space: nowrap;
+    overflow: hidden;
+}
+
+.autocomplete-selected {
+    background: #F0F0F0;
```
flodolo: Same here

web/style/transvision.css
```diff
@@ -578,6 +578,37 @@ fieldset {
...
+}
+
+.autocomplete-suggestions strong {
+    font-weight: normal;
+    color: #3399FF;
```
flodolo: Same here (lowercase)

flodolo:
I haven't tested the code, but I think we should have the new JavaScript code inside of `main_search.js`, not `base.js`.

Besides a few nits, this works really well, but I find confusing that I only get suggestions from English. Not sure how much can be done to add a second API query and show both results.

app/views/templates/base.php
pascalchevrel: ok

pascalchevrel:
If you switch locales temporarilly with the locale switcher widget, you can get results in your language.

flodolo:
I'm not sure how intuitive it is, but I also don't think having 2 requests for each character makes sense (the alternative is adding an API call just for suggestions, which might be used also for a searchplugin).

Anyhow, it's an r+ from me :+1:

Issue #554, opened 2015-12-26
Add API call for suggestions


flodolo:
Add an API call that returns suggestions from both source and target locale.

It can be used to provide suggestions while typing, but also for an external searchplugin.

Already have a working WIP, still need to add tests and check if it can be optimized a bit.

PR #555, opened 2015-12-26
Add API for suggestions

flodolo:
Fix issue #554 

I confess I haven't looked much into the API part of Transvision, but I'm not sure why we're returning instead of assigning $json in models ([example](https://github.com/mozfr/transvision/blob/master/app/models/api/translation_memory.php#L48))


app/models/api/suggestions.php
```diff
@@ -0,0 +1,58 @@
...
+
+// Regex options (not currenty used)
+$delimiter = '~';
+$whole_word = isset($check['whole_word']) ? '\b' : '';
+$case_sensitive = isset($check['case_sensitive']) ? '' : 'i';
+$regex = $delimiter . $whole_word . $initial_search . $whole_word .
+         $delimiter . $case_sensitive . 'u';
+
+foreach ($repositories as $repository) {
+    // Loop through all repositories searching in the source locale
```
pascalchevrel: I see you moved the comment inside the loop, the interest of having it before foreach is that you can read it and follow the file logic by reading comments if you fold all the code (in Atom, Edit/Fold/Fold All)

app/models/api/suggestions.php
```diff
@@ -0,0 +1,58 @@
...
+
+    return $value;
+};
+
+$json = ShowResults::getSuggestionsResults(
+    $source_strings_merged,
+    $target_strings_merged,
+    $initial_search,
+    $get_option('max_results')
+);
```
pascalchevrel: For this file, there is some code duplication with translation_memory.php,  not sure how to avoid that so not a blocker.

app/classes/Transvision/API.php
```diff
@@ -240,6 +266,9 @@ private function isValidServiceCall($service)
                 }
 
                 break;
+            case 'suggestions':
+            // Use the same settings as 'tm'
+            // ex: /api/v1/suggestions/release/en-US/fr/string/Home%20page/?max_results=3&min_quality=80
```
pascalchevrel: min_quality parameter should be removed from the example

app/classes/Transvision/ShowResults.php
```diff
@@ -35,6 +35,61 @@ public static function getTMXResults($entities, $array_strings)
     }
 
     /**
+     * Return an array of suggestions from our Translation Memory API. Unlike
+     * other services, we return values from both source and target language.
```
pascalchevrel: languages

app/models/api/suggestions.php
flodolo: Good point, I never thought about it. In this specific case I moved it inside because there are two cycles, but I think it make sense to have one outside.

pascalchevrel:
LGTM with a couple of nits, can you also update the Wiki page about the API once this is merged?

flodolo:
@pascalchevrel what about returning vs just assigning $json in controllers?

pascalchevrel:
I have a preference for returning because it makes sure we don't add any extra code in the future after $json that would have an unintended impact, but that's a personal preference I think.

flodolo:
OK, makes sense. I think I'll use that for consistency then.

flodolo:
Update Wiki
https://github.com/mozfr/transvision/wiki/JSON-API/_compare/3852a68d0cd943d71f4c435365353e8a1820c9c3...530d63cdd7b420daea79345ae365d3f3aa9e40b7

Issue #553, opened 2015-12-25
Update Travis file to exclude xdebug + add PHP 5.6 and 7


pascalchevrel:
I noticed in the travis log that php-cs-fixer is saying that:
> You are running php-cs-fixer with xdebug enabled. This has a major impact on runtime performance. 

Looking at the issues on Travis CI repo, this is also causing PHPUnit to be slow, so I guess it also impacts Atoum.

There is documentation to deactivate xdebug there:
https://docs.travis-ci.com/user/languages/php/#PHP-5.6-and-below

We should also replace PHP 5.5 by 5.6 (since we will move to Debian 8 soon) and add 7.0 to the matrix since I hope we will move to that in 2016.

@SkySymbol want to do this one?

SkySymbol:
Yes, you can assign me that issue. Switching to 7.0 will be a huge work :-)

pascalchevrel:
thanks

PR #556, opened 2015-12-26
disable xdebug, add PHP 5.6 and 7.0

SkySymbol:


.travis.yml
```diff
@@ -3,12 +3,16 @@ language: php
...
+  - 7.0
 before_script:
   ## Composer
   - composer self-update
   - composer install --no-interaction --prefer-source
-  ## travis configuration
+  ## Travis configuration
   - cp app/config/config.ini-travis app/config/config.ini
+  ## Disabling xdebug
+  - phpenv config-rm xdebug.ini
```
flodolo: I think you need to put it at the beginning of the `before_script` block

SkySymbol: Done, thank you for the review

flodolo:
Weird, the warning went away for ```php-cs``` but not for ```composer --version```, I assumed it was a order issue but it's not.

flodolo:
Let's see if @pascalchevrel has ideas. Also not sure if he wants to keep 5.5 based on #553

pascalchevrel:
yes, 5.5 should be removed.

PR #559, opened 2015-12-28
Clean up PHP warnings

flodolo:
Noticed them while checking my test server


flodolo:
Updated: As far as I can tell the first cycle to set `$results` is completely unused.

Issue #549, opened 2015-12-23
Remove 'wildcard' search option


pascalchevrel:
The reasons are:
- I don't understand myself how it is supposed to work and I think we actually broke it a long time ago and nobody noticed, so basically noibody uses it ;)
- I am rewriting the whole search process with a class that also takes care of defining the regex and I would like to remove the implementation of that feature since it just reproduces the current broken behaviour
- it would simplify our option choices with the ones that are really useful

Thoughts?

pascalchevrel:
@flodolo @TheoChevalier @SkySymbol  pinging active people :)

pascalchevrel:
btw, I would also rename "Whole word" ad "Contiguous words", we use the regex wording for the feature and I think nobody gets what it actually means

flodolo:
Absolutely 'yes' for wildcard, I'm not convinced about 'Whole word': I use it exactly to search for the entire word, e.g. [cookie](https://transvision.mozfr.org/?recherche=cookie&repo=aurora&sourcelocale=en-US&locale=it&search_type=strings&whole_word=whole_word) (i.e. I don't want 'cookies').

'Contiguous words" sounds like a search for "open window" should only return a search with strings that contain "open window", in that order.

TheoChevalier:
Agreed for wildcard, never used it not understood it.
Regarding Whole word, I’m fine to change it if we find something clearer, but maybe we could also use a tooltip to give more detail? Like "Only keep results containg those words"

pascalchevrel:
ok, so let's scrap wildcard and brainstorm on the right wording for "whole word" as we have several use cases, like the one by flod with a single word and the one I have when I want to return strings that have contiguous words seen as a single string, ex:
https://transvision.mozfr.org/?recherche=Private+Browsing&repo=release&sourcelocale=en-US&locale=fr&search_type=strings&whole_word=whole_word

I like the tooltip ideas that we could add to all options actually, I would put that in a separate issues as a good first bug.

SkySymbol:
+1 for the wildcard search option, but I disagree for renaming the whole word feature, as it seems to be the correct appellation. See for example  [this post](http://stackoverflow.com/questions/1751301/regex-match-entire-words-only) on Stackoverflow.

pascalchevrel:
the problem with the naming is that the people that use Transvision don't know what a regex is, so the question is not what the right naming is technically, but what the right naming is to make the feature understandable by end-users.

pascalchevrel:
[...]

PR #562, opened 2015-12-30
Issue #549 remove wildcard search option

pascalchevrel:


flodolo:
Looks good, tested without issues.

Issue #320, opened 2014-07-10
Comparison view should allow comparing strings across any repositories


julen:
The comparison view currently only allows comparing strings across grouped repositories, i.e. [Gaia on the one hand](http://transvision.mozfr.org/gaia/) (`gaia-l10n`, `gaia-vxy`), and [the rest on the other hand](http://transvision.mozfr.org/channelcomparison/) (`central`,`aurora`, `beta`, `release`).

Since some strings that have already been localized in the products are being reused in Gaia (I can think of network-related strings, _This address is restricted_ for instance), it'd be great being able to compare strings across any repositories, so I could see how a string has been localized in `aurora` and `gaia-vxy` repositories.

pascalchevrel:
I am working on the "all repositories" search option.

julen:
Thank you @pascalchevrel, much appreciated!

PR #563, opened 2015-12-30
Issues #320: move all our regex logic into a unit tested class

pascalchevrel:
I only put in this branch the treatment of the search query.
Currently the Search class only implements methods to create the regex necesary for our various views. 

This class has unit tests, so the main benefits of this patch are that the definition of the regex in models is more intuitive (creation of an object with a fluent interface) and that it should be much harder  to regress our regex code in our models now.

### DIFF
--- a/app/classes/Transvision/Search.php
+++ b/app/classes/Transvision/Search.php
@@ -1,0 +1,228 @@
+<?php
+namespace Transvision;
+
+/**
+ * Search class
+ *
+ * Allows searching for data in our repositories using a fluent interface.
+ * Currently, only the regex part (definition of the search) is implemented.
+ * e.g.:
+ * $search = (new Search)
+ *     ->setSearchTerms('Bookmark this page')
+ *     ->setRegexWholeWords(true)
+ *     ->setRegexCaseInsensitive(true)
+ *     ->setRegexPerfectMatch(false);
+ */
+class Search
+{
+    /**
+     * The trimmed string searched, we keep it   as the canonical reference
+     * @var string
+     */
+    protected $search_terms;
+
+    /**
+     * The generated regex string updated dynamically via updateRegex()
+     * @var string
+     */
+    protected $regex;
+
+    /**
+     * Case sensibility of the regex
+     * @var string
+     */
+    protected $regex_case;
+
+    /**
+     * Consider the space separated string as a single word for search
+     * @var string
+     */
+    protected $regex_whole_words;
+
+    /**
+     * Only return strings that match the search perfectly (case excluded)
+     * @var boolean
+     */
+    protected $regex_perfect_match;
+
+    /**
+     * The search terms for the regex, these differ from $search_terms as
+     * they can be changed dynamically via setRegexSearchTerms()
+     * @var string
+     */
+    protected $regex_search_terms;
+
+    /**
+     * We set the default values for a search
+     */
+    public function __construct()
+    {
+        $this->search_terms = '';
+        $this->regex = '';
+        $this->regex_case = 'i';
+        $this->regex_whole_words = '';
+        $this->regex_perfect_match = false;
+        $this->regex_search_terms = '';
+    }
+
+    /**
+     * Store the searched string in $search_terms and in $regex_search_terms
+     *
+     * @param  string $string String we want to search for
+     * @return $this
+     */
+    public function setSearchTerms($string)
+    {
+        $this->search_terms = trim($string);
+        $this->regex_search_terms = $this->search_terms;
+        $this->updateRegex();
+
+        return $this;
+    }
+
+    /**
+     * Allows setting a new searched term for the regex.
+     * This is mostly useful when you have a multi-words search and need to
+     * loop through all the words to return results.
+     *
+     * @param  string $string The string we want to update the regex for
+     * @return $this
+     */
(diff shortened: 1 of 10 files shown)


app/classes/Transvision/Search.php
```diff
@@ -0,0 +1,228 @@
+<?php
+namespace Transvision;
+
+/**
+ * Search class
+ *
+ * Allows searching for data in our repositories using a fluent interface.
+ * Currently, only the regex part (definition of the search) is implemented.
+ * ex:
```
flodolo: Trivia: I noticed it while adding the suggestion API, is it a French abbreviation? Not sure how common it is in English

app/classes/Transvision/Search.php
```diff
@@ -0,0 +1,228 @@
...
+ * $search = (new Search)
+ *     ->setSearchTerms('Bookmark this page')
+ *     ->setRegexWholeWords(true)
+ *     ->setRegexCase('sensitive')
+ *     ->setRegexPerfectMatch(false);
+ */
+class Search
+{
+    /**
+     * The trimmed string searched, we keep that one as the canonical reference
```
flodolo: nit: that one -> it

app/classes/Transvision/Search.php
```diff
@@ -0,0 +1,228 @@
...
+    protected $regex_whole_words;
+
+    /**
+     * Only return strings that match the search perfectly (case excluded)
+     * @var boolean
+     */
+    protected $regex_perfect_match;
+
+    /**
+     * The search terms for the regex, those differ from $search_terms as
```
flodolo: nit: those -> these

app/classes/Transvision/Search.php
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Set the regex case sensibility.
+     *
+     * @param  boolean $flag 'sensitive' == '' in a regex
+     * @return $this
+     */
+    public function setRegexCase($flag)
+    {
+        $this->regex_case = (boolean) $flag ? '' : 'i';
```
flodolo: This is quite confusing: the function is called `setRegexCase`, but it should actually be `setRegexCaseSensitive`.
Proof is that calling `$this->setRegexCase('insensitive')` set the regex as case sensitive.

I would also update the example at the beginning of the class with an explicit true or false.

app/classes/Transvision/Search.php
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * We set the default values for a search
+     */
+    public function __construct()
+    {
+        $this->search_terms = '';
+        $this->regex = '';
+        $this->regex_case = 'i';
+        $this->regex_whole_words = '';
```
flodolo: I'd be more explicit and set `false` for these two (even if we cast them to boolean later).

flodolo:
It definitely looks good to me. 

Only one possible concern: we end up calling the updateRegex method a scary amount of times, often to not change anything (I realized it while logging the regex). 

The only alternative I can think of it to remove the call from set methods and call it explicitly at the end

```
$search = (new Search)
    ->setSearchTerms(Utils::cleanString($_GET['recherche']))
    ->setRegexWholeWords($check['whole_word'])
    ->setRegexCase($check['case_sensitive'])
    ->setRegexPerfectMatch($check['perfect_match'])
    ->updateRegex();
```
The only downside is remembering that we need to call updateRegex at the end.

app/inc/search_options.php
```diff
@@ -40,18 +40,13 @@
...
-$whole_word     = $check['whole_word']     ? '\b' : '';
-$delimiter      = '~';
-$main_regex     = $delimiter . $whole_word . $my_search . $whole_word . $delimiter . $case_sensitive;
+// Define our regex
+$search = (new Search)
+    ->setSearchTerms(Utils::cleanString($_GET['recherche']))
+    ->setRegexWholeWords($check['whole_word'])
+    ->setRegexCase($check['case_sensitive'])
+    ->setRegexPerfectMatch($check['perfect_match'])
+;
```
flodolo: I find the semicolon on a new line quite strange

app/classes/Transvision/Search.php
pascalchevrel: I can change it to e.g.

app/classes/Transvision/Search.php
pascalchevrel: that means making updateRegex() more complex as we need then to change them back to strings there

flodolo: Never mind, this was a dumb comment. These two store the regex piece, not a status, so they're correct as they are.

pascalchevrel:
I think it's more risky to have to manually add updateRegex() everytime we change a setter than to have it called automatically. Furthermore, it will not reduce the number of calls significantly, only a handful of times, because it is called in two places only, at the initial definition of the regex (we would save like 3 calls to updateRegex()) and in loops when we have to search for each word, in that case there would be no saving at all because we would have to add a call to updateRegex() anyway if it's not called automatically.

flodolo:
I agree on the risk, the number of request would be always 1/3. It doesn't seem much, but it's 3+3*character typed slowly in the search bar+3 for the resulting page: that's 10 vs 30 for "window". Still, probably not worth the risk of thinking the regex is updated while it's actually not.

PR #564, opened 2015-12-31
Don't display suggestions for entities search

flodolo:


pascalchevrel:
I tried locally and it only works if you are already looking for entities with a search such as:
http://localhost:8082/?recherche=&repo=release&sourcelocale=en-US&locale=fr&search_type=entities

If you change the value of the select box for search_type, it doesn"t work, either we don't get suggestions for strings if we were on an entities search, or we get suggestions for entities if we were on a strings search.

flodolo:
True :-\

Let me think if there's another way to fix it.

flodolo:
We already have a listener on the onChange of the select, should be easy to fix.

flodolo:
Updated. It took a lot more thinking than I expected.

PR #567, opened 2016-01-01
Issue #551: fix autocomplete in Chrome

pascalchevrel:
Chrome does not take the 'params' option into account to update the query with this error message:

Uncaught TypeError: Cannot assign to read only property 'query' of max_results=10b.getSuggestions @ jquery.autocomplete.min.js?v=unknown.dev:8b.onValueChange @ jquery.autocomplete.min.js?v=unknown.dev:8b.onKeyUp @ jquery.autocomplete.min.js?v=unknown.dev:8(anonymous function) @ jquery.autocomplete.min.js?v=unknown.dev:8m.event.dispatch @ jquery.min.js?v=unknown.dev:4r.handle @ jquery.min.js?v=unknown.dev:4

Appending the max_results paramater directly to the return value of serviceurl works in both Firefox and Chrome

### DIFF
--- a/web/js/main_search.js
+++ b/web/js/main_search.js
@@ -24,7 +24,7 @@ $(document).ready(function() {
                         + $('#target_locale').val() + '/'
                         + query + '/';
                 },
-                params: 'max_results=10',
+                params: { max_results: 10 },
                 minChars: 2,
                 transformResult: function(response) {
                     var data = JSON.parse(response);


flodolo:
Wow, that error message is not helpful. I think I found the issue in a [random post on SO](http://stackoverflow.com/questions/26602063/additional-params-in-jquery-autocomplete): params needs to be an object (completely undocumented).

This seems to work on both Firefox and Chrome:
```
params: { max_results: 10 },
``
I noticed the plugin also sends an extra parameter 'query=xxx', tried to find a way to avoid that but failed.

PR #570, opened 2016-01-04
Version 4.0

pascalchevrel:


app/models/changelog_data.php
```diff
@@ -46,9 +47,119 @@
     '3.8.1' => '2015-06-24',
     '3.9'   => '2015-10-08',
     '3.10'  => '2015-11-13',
+    '4.0'   => '2016-01-04',
 ];
 
 $changelog = [
+    '4.0' => [
+        [
+            'message' => ['Propose search suggestions as your type in the search field'],
```
flodolo: your->you

app/models/changelog_data.php
```diff
@@ -46,9 +47,119 @@
...
+        ],
+        [
+            'message' => ['Stop hard-coding version information in the application.'],
+            'section' => ['developers'],
+            'type'    => ['better'],
+            'commit'  => ['b7dbe66c4837ef557a40c57ee5baab86dff2d353'],
+            'authors' => ['flod'],
+        ],
+        [
+            'message' => ['Update Travis CI configuration toexclude xdebug and add PHP 5.6 and 7 to the testing matrix'],
```
flodolo: to exclude

app/models/changelog_data.php
```diff
@@ -46,9 +47,119 @@
...
+            'authors' => ['pascal'],
+        ],
+        [
+            'message' => ['Upgraded OS from Debian 7 to 8. Transvision minimum PHP requirement moved from 5.4 to 5.6'],
+            'section' => ['developers'],
+            'type'    => ['better'],
+            'authors' => ['geb'],
+        ],
+        [
+            'message' => ['Host OpenSans font ourselves to improve page load tome instead of loading it from mozilla.org'],
```
flodolo: time

app/models/changelog_data.php
```diff
@@ -46,9 +47,119 @@
...
+    '4.0' => [
+        [
+            'message' => ['Propose search suggestions as your type in the search field'],
+            'section' => ['end_user'],
+            'type'    => ['new'],
+            'issues'  => [551],
+            'authors' => ['pascal', 'flod'],
+        ],
+        [
+            'message' => ['Create a <a href="/rss/">RSS</a> feed for release notes'],
```
TheoChevalier: I think we need a full link for external RSS readers

app/models/changelog_data.php
```diff
@@ -46,9 +47,119 @@
...
+        ],
+        [
+            'message' => ['Update results count when filtering search results by component'],
+            'section' => ['end_user'],
+            'type'    => ['bug'],
+            'commit'  => ['f26b5fb58047ba8c460ff6bd41dabb461b18a91d'],
+            'authors' => ['flod'],
+        ],
+        [
+            'message' => ['Stop hard-coding version information in the application.'],
```
flodolo: Remove final period for consistency with other items.

app/models/changelog_data.php
flodolo: Why? It works fine for me

app/models/changelog_data.php
```diff
@@ -46,9 +47,119 @@
...
+            'authors' => ['pascal'],
+        ],
+        [
+            'message' => ['Transvision moved to a faster server with hardware redundancy sponsored by <a href="https://www.gandi.net/">Gandi</a>'],
+            'section' => ['developers'],
+            'type'    => ['better'],
+            'authors' => ['pascal'],
+        ],
+        [
+            'message' => ['Upgraded OS from Debian 7 to 8. Transvision minimum PHP requirement moved from 5.4 to 5.6'],
```
TheoChevalier: I see you updated the Wiki, but we will need to update the readme file as well :) https://github.com/mozfr/transvision#dependencies

app/models/changelog_data.php
TheoChevalier: If you read it from Thunderbird, I think it won’t work (isn’t it the reason why we updated all the relative links in this file?). If it works then we can probably revert the change we made

app/models/changelog_data.php
TheoChevalier: Didn’t see it’s done in this PR :)

app/models/changelog_data.php
pascalchevrel: I just tried and it works without the domain.

TheoChevalier: Ok sweet, so we can s/https:\/\/transvision.mozfr.org// (Or I can do a separate PR if you want)

PR #582, opened 2016-01-07
Add functional tests for public pages

SkySymbol:


tests/functional/pages.php
```diff
@@ -0,0 +1,65 @@
...
+$pid = $output[0];
+
+// Pause to let time for the dev server to launch in the background
+sleep(3);
+
+$paths = [
+    ['', 200, 'You are looking at', 'Type your search here'],
+    ['3locales/', 200, 'One source locale', 'Type your search here'],
+    ['accesskeys/', 200, 'Check your access keys', 'Label entity'],
+    ['changelog.php/', 200, 'You are looking at', 'Type your search here'],
```
flodolo: I think we should remove this, and also remove it from urls.php (I'll file a new issue for that).
It's probably a remnant of the past (before nice URLs), and it also generate warnings

tests/functional/pages.php
```diff
@@ -0,0 +1,65 @@
...
+    ['channelcomparizon/', 200, 'Compare strings from channel to channel', 'Locale'],
```
flodolo: Let's test for `channelcomparison`. Also, as second value I'd use 'Key'

flodolo:
There's one url missing: `/string/`

Besides that, I think it's a good start. We don't currently have redirects or a 404 page, so I think it makes sense to test only 200 responses.

SkySymbol:
Thanks for reviewing :-)

flodolo:
You're welcome, and thanks for the patch.

@TheoChevalier 
Mind doing a second pass?

tests/functional/pages.php
```diff
@@ -0,0 +1,63 @@
...
+    ['channelcomparizon/', 200, 'Compare strings from channel to channel', 'Key'],
+    ['channelcomparison/', 200, 'Compare strings from channel to channel', 'Key'],
+    ['credits/', 200, 'Transvision 1.0 was created', 'Transvision is hosted on the MozFR server'],
+    ['downloads/', 200, 'Select which strings', 'Generate the TMX'],
+    ['gaia/', 200, 'Translation Status', 'How many strings are translated'],
+    ['news/', 200, 'Version 4.0', 'End user visible changes'],
+    ['productization/', 200, 'Show productization', 'firefox'],
+    ['showrepos/?locale=en-GB', 200, 'Health Status for en-GB', 'General metrics'],
+    ['stats/', 200, 'Repository status overview', 'Status estimate'],
+    ['string/?entity=apps/sms/ssms.properties:home&repo=gaia', 200, 'All translations for this string', 'supported_locales'],
```
TheoChevalier: typo: sms.properties.
It works because the title is displayed even if the data is not found. But I think it would make sense to test data that is available both locally and in Travis (https://github.com/mozfr/transvision/blob/master/tests/testfiles/TMX/ ). We could check for instance that for the string bookmarkResultLabel in mozilla-central, the translation Marque-page is displayed in the page, like that:

['string/?entity=browser/chrome/browser/places/places.properties:bookmarkResultLabel&repo=central', 200, 'All translations for this string', 'supported_locales', 'Marque-page'],

This should work in Travais as well.

TheoChevalier:
Otherwise, looks good to me :)

SkySymbol:
That makes sense, thank you for the review

flodolo:
Looks good, thanks :+1:

PR #588, opened 2016-01-14
Clean up public urls

### LINKED ISSUE #583: Clean up available URLs
While looking at the functional tests for pages, I noticed some strange URLs in the app
* `changelog.php`: I assume it was maintained for retro-compatibility, but I don't think we should keep it (checking access.log on the prod server could confirm that)
* `repocomparizon`, `repocomparison`: empty page
* `channelcomparizon`: we should keep only `channelcomparison`

Also note that most of these views generate warnings in the log.

SkySymbol:
Fix #583, removing unnecessary urls.

### DIFF
--- a/app/inc/urls.php
+++ b/app/inc/urls.php
@@ -3,12 +3,8 @@
 $urls = [
     '/'                 => 'root',
     'news'              => 'changelog',
-    'changelog.php'     => 'changelog',
     'rss'               => 'rss',
     'stats'             => 'stats',
-    'repocomparizon'    => 'repocomp',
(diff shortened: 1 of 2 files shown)


TheoChevalier:
Looks like you need to update the functional tests and remove /channelcomparizon/ :)
(And you already know where it is :p)

SkySymbol:
Right, my bad

TheoChevalier:
Thank you @SkySymbol ! :beers: 
If @flodolo has no comment, it looks good to me

flodolo:
It looks good, thanks.

PR #589, opened 2016-01-15
Add |./start.sh -tests| option

TheoChevalier:


TheoChevalier:
This doesn’t scale, but I can’t think of something simpler https://github.com/mozfr/transvision/pull/589/files#diff-d0b2ef35f4d7cb0284dd5cf566b62f51R13

start.sh
```diff
@@ -3,23 +3,38 @@
...
 SERVER="localhost:8082"
 
+if [ $# -gt 0 ] && [ $1 = '-help' ]
+then
+    echo "Usage: start.sh (PHP web server will be listening to localhost:8082)"
+    echo "Usage: start.sh -remote (PHP web server will be listening to 0.0.0.0:8082 and accessible from the outside)"
+    echo "Usage: start.sh -tests (Run all the unit and fonctional tests and check PHP syntax with PHP-cs-fixer)"
+    echo "Additional parameters will be ignored."
+    exit 1
+elif [ $1 != '-remote' ] && [ $1 != '-tests' ] && [ $1 != '-tests-server' ]
```
TheoChevalier: This doesn’t scale, but I can’t think of something simpler

flodolo: Possible alternative with array (I don't think we should support `-tests-server` at all.

```bash
SUPPORTED_COMMANDS=(
    '-help'
    '-remote'
    '-tests'
)

if [ $# -gt 0 ]
then
    supported_command=false
    for command in ${SUPPORTED_COMMANDS[@]};
    do
        if [ "${command}" == "$1" ];
        then
            supported_command=true
            break
        fi
    done
    if [ "${supported_command}" = false ];
    then
        echo "Unknown parameter \`${1}\`. Type \`start.sh -help\` to know what the valid parameters are."
        exit 1
    fi
fi

if [ $1 = '-help' ]
then
    echo "Usage: start.sh (PHP web server will be listening to localhost:8082)"
    echo "Usage: start.sh -remote (PHP web server will be listening to 0.0.0.0:8082 and accessible from the outside)"
    echo "Usage: start.sh -tests (Run all the unit and fonctional tests and check PHP syntax with PHP-cs-fixer)"
    echo "Additional parameters will be ignored."
    exit 1
fi
```

start.sh
```diff
@@ -3,23 +3,38 @@
 # Use 'start.sh -remote' if you want to access the server from another machine (or a VM)
 SERVER="localhost:8082"
 
+if [ $# -gt 0 ] && [ $1 = '-help' ]
+then
+    echo "Usage: start.sh (PHP web server will be listening to localhost:8082)"
+    echo "Usage: start.sh -remote (PHP web server will be listening to 0.0.0.0:8082 and accessible from the outside)"
+    echo "Usage: start.sh -tests (Run all the unit and fonctional tests and check PHP syntax with PHP-cs-fixer)"
```
flodolo: typo: functional

start.sh
TheoChevalier: Nice trick with the array :) Regarding -tests-server, Pascal wanted to be able to run tests while Transvision is running on 8082, hence the use of 8083 with -tests-server

flodolo: Never mind, I didn't realize it was an option later on, somehow my brain thought it was the concatenation of two options...

flodolo: Note: we should probably document it in the help

TheoChevalier: I didn't want to document it as it's internal use only, but if you think it should, let's add it.

flodolo: Yeah, I think it makes sense.

flodolo:
It could be merged for me. We could also simplify the last if series with a case
```
case "$1" in
    -remote)
        SERVER="0.0.0.0:8082"
        ;;
    -tests)
        php ./vendor/atoum/atoum/bin/atoum -d tests/units/
        php ./tests/functional/api.php
        php ./tests/functional/pages.php
        vendor/bin/php-cs-fixer --diff --dry-run -v fix
        exit 1
        ;;
    -tests-server)
        SERVER="localhost:8083"
        ;;
esac
```

TheoChevalier:
Added the case, thanks flod ;)

flodolo:
Thanks, good to merge for me :bear:

PR #590, opened 2016-01-19
Add a meta description tag

### LINKED ISSUE #579: Add a meta description tag
We currently don’t have a description meta tag, so the site description in search engine results is quite bad.
We should add a nice <meta name="description" content="Foo bar"> tag in pages header to simply describe the tool.

SkySymbol:
Fix #579. I've used the description from the Github repo

### DIFF
--- a/app/views/templates/base.php
+++ b/app/views/templates/base.php
@@ -75,6 +75,7 @@ if (file_exists(CACHE_PATH . 'lastdataupdate.txt')) {
 } ?><?= $title_productname ?></title>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="description" content="Tranvision is a tool used by the Mozilla Community to search translations extracted from products and websites. It also provides API access to translation memory, and specific QA features">
 <?php foreach ($css_include as $css_file):?>
     <link rel="stylesheet" href="/style/<?= $css_file . $cache_bust ?>" type="text/css" media="all" />
 <?php endforeach?>


app/views/templates/base.php
```diff
@@ -75,6 +75,7 @@
 } ?><?= $title_productname ?></title>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="description" content="Transvision is a Web application targeting the Mozilla localization community, created and maintained by the French Mozilla community">
```
pascalchevrel: We should be more specific on what it is exactly.
* We don't need to mention that the French community did it
* We need to explain that this Web appllication is a translation memory tool

Ex:
Transvision is a translation memory application for the Mozilla localization community. It indexes translations from Mozilla software and websites in over 80 languages and allows searching for strings for QA purposes, downloading the data in TMX format among other features. It also provides a public json API allowing to retrieve translations and metadata for reuse. 

(can be improved, please check for possible typos too ;) )

flodolo: I think we might need to shorten it, usually a description should be around 160 characters or less (unless we don't want to bother about Google)
https://moz.com/blog/i-cant-drive-155-meta-descriptions-in-2015

flodolo: One possible suggestion before I forget it: «Tranvision is a tool used by the Mozilla Community to search translations extracted from products and websites. It also provides API access to translation memory, and specific QA features.»

The second sentence might get truncated, but it would still make sense.

SkySymbol:
Thanks

pascalchevrel:
LGTM merging, thanks :)

Issue #591, opened 2016-01-19
Don't cache a whole repo string in Utils::getRepoStrings() 


pascalchevrel:
While looking at this api call and seeing how we can make it faster:
https://transvision.mozfr.org/api/v1/tm/global/en-US/fr/Cancel/

I noticed that the page renders much faster and wjth less memory if I disable the cache:
With cache:
[Tue Jan 19 16:57:56 2016] Memory peak: 38535168 (36.75MB)
[Tue Jan 19 16:57:56 2016] Elapsed time (s): 0.6915
Whithout cache:
[Tue Jan 19 16:58:15 2016] Memory peak: 23592960 (22.5MB)
[Tue Jan 19 16:58:15 2016] Elapsed time (s): 0.3894

I digged into it and I see that I think I made a conceptual error in the function getRepoStrings()

This function will *always* return the full included PHP file (```{$locale}/cache_{$locale}_{$repository}.php```) so we don't need to put that file in cache because that means that everytime we need to work on the full set of strings we serialize/unserialize it.

That is true will all views using getRepoStrings() but it  is particularily noticeable on the global api one because it needs to load strings for all reporitories so we spend 30% of the page time view in unserialize().

PR #592, opened 2016-01-19
Issue #591: Don't cache the whole repository for a locale in getRepoS…

pascalchevrel:
…trings()

### DIFF
--- a/app/classes/Transvision/Utils.php
+++ b/app/classes/Transvision/Utils.php
@@ -232,15 +232,17 @@ class Utils
         $locale = Project::getLocaleInContext($locale, $repository);
 
         $file = TMX . "{$locale}/cache_{$locale}_{$repository}.php";
-
-        if (! $tmx = Cache::getKey($file)) {
-            if (is_file($file)) {
-                include $file;
-                Cache::setKey($file, $tmx);
-            }
+        if (! is_file($file)) {
+            return [];
         }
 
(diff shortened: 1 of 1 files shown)


app/classes/Transvision/Utils.php
```diff
@@ -233,13 +233,12 @@ public static function getRepoStrings($locale, $repository)
...
-                include $file;
-                Cache::setKey($file, $tmx);
-            }
+        if (! is_file($file)) {
+            return [];
         }
 
+        include $file;
+
         return $tmx !== false ? $tmx : [];
```
flodolo: Does this part need updating? In which case $tmx could be false?

pascalchevrel: if the file is empty I guess (created  by the script but nothing got written into it). Very very unlikely, I am OK to simplify it.

flodolo: If the file is broken we should probably get a warning for `$tmx` not being defined (before it would have been at least false because of Cache::getKey), so I think it makes sense to test if it's set?

pascalchevrel: using error_log() to log to the apache log is ok for you?

flodolo: Yes. Transvision's log is pretty clean right now, so errors should be highly visible.

Issue #595, opened 2016-01-19
Improve perf of tm api with global search


pascalchevrel:
We have already improved the perf of this view with the fix of getRepoStrings() but I found ways to make it faster and also simplify the getTranslationMemoryResults() method.

pascalchevrel:
perf improved between x2 and x4 depending on cases, should be good enough.

PR #596, opened 2016-01-19
Issue #595 faster Translation Memory (tm) API

pascalchevrel:
```
/api/v1/tm/global/en-US/fr/Cancel/
```
Before:
[Wed Jan 20 00:04:54 2016] Memory peak: 23592960 (22.5MB)
[Wed Jan 20 00:04:54 2016] Elapsed time (s): 0.3324

After:
[Wed Jan 20 04:12:04 2016] Memory peak: 8388608 (8MB)
[Wed Jan 20 04:12:04 2016] Elapsed time (s): 0.145

The gain is ~x2.2

If the search is for a string that exists only in one small repository, mozilla.org for example, the gains are even bigger (~x4):

```
/api/v1/tm/global/en-US/fr/Foundation pledge/
```
Before:
[Wed Jan 20 01:04:55 2016] Memory peak: 23330816 (22.25MB)
[Wed Jan 20 01:04:55 2016] Elapsed time (s): 0.2898

After:
Wed Jan 20 04:14:04 2016] Memory peak: 7077888 (6.75MB)
[Wed Jan 20 04:14:04 2016] Elapsed time (s): 0.0796

### DIFF
--- a/app/models/api/translation_memory.php
+++ b/app/models/api/translation_memory.php
@@ -11,6 +11,7 @@ $get_option = function ($option) use ($request) {
 
     return $value;
 };
+
 $repositories = ($request->parameters[2] == 'global')
     ? Project::getRepositories()
     : [$request->parameters[2]];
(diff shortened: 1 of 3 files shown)


app/models/api/translation_memory.php
```diff
@@ -32,21 +33,51 @@
...
 
     foreach ($terms as $word) {
         $search->setRegexSearchTerms($word);
         $source_strings = preg_grep($search->getRegex(), $source_strings);
     }
 
-    $source_strings_merged = array_merge($source_strings, $source_strings_merged);
-    $target_strings_merged = array_merge($target_strings, $target_strings_merged);
+    /*
+        If we don't have any match for a repo, no need to do heavy calculation,
```
flodolo: nit: calculations

app/models/api/translation_memory.php
```diff
@@ -32,21 +33,51 @@
...
+        Utils::getRepoStrings($request->parameters[4], $repository),
+        $source_strings
+    );
+
+    /*
+        We are not interested in keeping duplicate strings that have
+        different keys because this API does not take into account the
+        frequency of matches but the similarity of the strings.
+    */
+    $target_strings = array_unique($target_strings);
```
flodolo: I might be lost, but what's the scenario here? ```$target_strings``` is set with new values for each repository, given that keys are in the form `browser/chrome/browser/aboutPrivateBrowsing.dtd:aboutPrivateBrowsing.info.cookies` in which cases would we have duplicated items?

pascalchevrel: array_unique() does not filter on keys but on values, so if we have multiple strings "Annuler' for example, with different keyx, only one Annuler string will be kept. Example in the doc http://php.net/manual/en/function.array-unique.php#refsect1-function.array-unique-examples

flodolo: Thanks, makes totally sense.

pascalchevrel:
good to go?

flodolo:
Yes, looks great :+1:

Issue #597, opened 2016-01-20
Truncate quality values in API


flodolo:
I don't think there's value in exposing so many decimals, e.g. source `quality: 42.857142857143`
https://transvision.mozfr.org/api/v1/tm/aurora/it/it/fin/?max_results=10

pascalchevrel:
The reason we don't truncate is to be fully compatible with Amagama that Pontoon also uses:
https://amagama-live.translatehouse.org/api/v1/en/cs/unit/Cancel

We can truncate but we need to check with matjaz first that it doesn't impact his code

flodolo:
To clarify: truncate as in exposing less decimals, not truncating to integers.

Given that quality could be 100 (or 40 in the example above), I wouldn't expect breakages.

pascalchevrel:
no problem with that, but let's check with matjaz first, I don't want anything to break on Pontoon :)

flodolo:
@mathjazz Do you see any issue with reducing the number of decimals for quality in our API response?

mathjazz:
r+

Pontoon uses `Math.round(quality)`, so it displays quality wihout decimals. Thanks for asking!

pascalchevrel:
Taking this one, I think 2 decimals is a good number

PR #598, opened 2016-01-20
Issue #597: round quality parameter in 'tm' API to 2 decimals

pascalchevrel:


flodolo:
Can you fix also getSuggestionsResults since it's basically the same code?

Besides that, everything looks good.

Issue #612, opened 2016-02-09
Improve speed and result quality for the translation memory api


pascalchevrel:
Not related to the Hackathon.

I found more perf improvements for the tm/global api and it also makes the code simpler. I also found out that with the simplification we now get a bit more results (not noticeable for big locales because looking for strings almost always returns a lot of results, but for locales that only do Gaia for example, a search that would yield 3 results now would return maybe 5 results).

One thing I wonder is if we should also set a minimum quality threshold (currently the value is set to 0, that means we return all results even if the quality is very low) to maybe a value of 50. It has no impact on the speed but maybe a higher default value would be a good thing.

PR #613, opened 2016-02-09
issue #612: Improve speed and result quality for global translation memory api

pascalchevrel:
* For a global search, don't loop through all repositories, only those a locale supports (major perf gain for small locales)
* Simplify ShowResults::getTranslationMemoryResults(), send a prepared array without entities as keys
* Ignore empty localized strings in search results (happens when the entity is in the repo but empty)

For locales that do everything, this has no perf impact, for most locales that will give a small perf boost (most locales don't do mozilla-central), for locales that have less repositories (just started and only have Aurora for example, or only translating Gaia, or not translating Firefox for iOS, the perf gain can be significant).
For Guarani for example, request times are divided by 5 on my machine compared to our current code. Same for memory use.

### DIFF
--- a/app/models/api/translation_memory.php
+++ b/app/models/api/translation_memory.php
@@ -13,11 +13,11 @@ $get_option = function ($option) use ($request) {
 };
 
 $repositories = ($request->parameters[2] == 'global')
-    ? Project::getRepositories()
+    ? Project::getLocaleRepositories($request->parameters[4])
     : [$request->parameters[2]];
 
-$source_strings_merged = [];
-$target_strings_merged = [];
+// This is the filtered data we will send to getTranslationMemoryResults()
+$output = [];
 
 // The search
 $initial_search = Utils::cleanString($request->parameters[5]);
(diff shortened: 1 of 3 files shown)


app/models/api/translation_memory.php
```diff
@@ -49,35 +49,23 @@
...
-    $target_strings_merged += $target_strings;
+    $target_strings = Utils::getRepoStrings($request->parameters[4], $repository);
 
+    foreach ($source_strings as $key => $value) {
+        if (isset($target_strings[$key]) && ! empty($target_strings[$key])) {
+            $output[] = [
+                $value,
+                $target_strings[$key],
+            ];
+        }
```
flodolo: It works fine, my only nit would be to use explicit indexes to make the code clearer in the class (`$set['source']` vs `$set[0]`). Not sure if there's a memory impact from that.
```PHP
            $output[] = [
                'source' => $value,
                'target' => $target_strings[$key],
            ];
```

flodolo:
Besides one note, it looks good to me.

app/models/api/translation_memory.php
pascalchevrel: I am testing and it does have an impact on perf (for French, requests that were taking 0.17s are now taling 0.2s), I think it is because creating a multidimensionnal array with associative keys is more expensive than one without. That said, I can put inline comments to mention which one is the source and which one is the target string

flodolo: Thanks for checking. I guess a comment explaining the structure of the array would work.

PR #614, opened 2016-02-11
Refactor Python script to create TMX for products

flodolo:


flodolo:
The main reason for this PR is that I couldn't fully understand some parts of the code, so I decided to rewrite it in a [separate repository](https://github.com/flodolo/tmx_maker/) with tests (happy to move the repo to the MozFR organization).

## Refactor
The first step was to make the code testable and refactor it in smaller functions. During the refactor I kept comparing the output with the old script, making sure there were no differences.

## Tests and behavior changes
Then I started adding tests to document the current behavior, and found some issues.

### Gaia repositories
For standard products we have an hardcoded list of directories to traverse, for Gaia we created a list of directories in reference and source, intersected them to analyze only folders available in reference.

I don't think that's necessary: `get_strings` is already a recursive function, we can start traversing from the root. Note that the previous code would also ignore files in the root.
https://github.com/flodolo/tmx_maker/commit/2e9035d589e6838ea0eee2f7d76b86af6cadfd8b

### Escaping
That's the part I could understand less, and I believe it was also wrong in dealing with single quotes already escaped. With the existing code, `This is a 'test'`and `This is a \'test\'` would generate the same array item.
https://github.com/flodolo/tmx_maker/commit/2f3959252ab5021b3d191320e456fb018129bd3e

If I compare French and Italian on Aurora before and after this PR there are no changes, for German the only diff is
```DIFF
19830c19830
< 'mobile/android/base/android_strings.dtd:home_add_panel_installed' => '\'&formatS;\' zum Startbildschirm hinzugefügt',
---
> 'mobile/android/base/android_strings.dtd:home_add_panel_installed' => '\\\'&formatS;\\\' zum Startbildschirm hinzugefügt',
``

Which seems correct compared to what we were doing.

**Comment by flodolo** [comment]
I realized it's relatively easy to compare the output before and after:
* Set glossaire.sh to not update sources but create TMX.
* Don't create TMX for Gaia (files are identical, but the order of keys is different).
* Compare the two TMX folders (master vs this branch)

http://pastebin.com/s7Yn0ARX
All the differences are escaped quotes that the current script doesn't manage correctly.

**Comment by flodolo** [comment]
Updated link with diff syntax highlighting
http://pastebin.com/nvJLreqT

**Comment by pascalchevrel** [comment]
I don't find differences before and after except the escaped strings which I am ok with. The fact that the strings in the gaia repos are no longer categorized by keys in the generated php array worries me a bit, I think it would work to do this line 154, that would generate sorted arrays for all repos, which can be useful when you open the file directly:
```python
    return sorted(tmx_content)
```

app/scripts/tmx_products.py
```diff
@@ -0,0 +1,192 @@
...
+    parser.add_argument('locale_code', help='Locale language code')
+    parser.add_argument('reference_code', help='Reference language code')
+    parser.add_argument('repository', help='Repository name')
+    args = parser.parse_args()
+
+    dirs = create_directories_list(
+        args.reference_repo, args.locale_repo, args.repository
+    )
+    tmx_content = create_tmx_content(
+        args.reference_repo, args.locale_repo, dirs)
```
pascalchevrel: parenthesis on a new line?

flodolo: It doesn't work in Python, you need a parenthesis to go on a new line (this is also the result of formatting for PEP8) :-\

flodolo: (or did you mean the closing parenthesis?)

pascalchevrel: yes, the closing parenthesis, for consistency with the definition of 'dirs' above

flodolo:
I'm a bit afraid by sorting the array for standard products (number of locales and size of the array), but I've added sorting for Gaia (left in a separate commit for simplify review)

flodolo:
Uhm, let me double check, it doesn't seem to have any effect.

flodolo:
OK, turns out the TMX created by the current script is in completely random order (can you confirm?), so I still get a huge diff if I order with the new one. Does it make sense to order it at this point?

pascalchevrel:
yes, I think it's random order. My thinking is that if the array is pre-ordered, it might have a positive perf impact when we do searches for entities.

flodolo:
OK. Let me try to order all repos and see if it has significant performance issues

flodolo:
Added sorting for all repos. The impact seems to be about 200ms (+2.5s to update Italian, so 12 repositories).

app/scripts/tmx_products.py
```diff
@@ -0,0 +1,194 @@
...
+import sys
+from ConfigParser import SafeConfigParser
+
+# Get absolute path of ../config from the current script location (not the
+# current folder)
+config_folder = os.path.abspath(
+    os.path.join(os.path.dirname(__file__), os.pardir, 'config'))
+
+# Read Transvision's configuration file from ../config/config.ini
+# If not available use default a /storage folder so store data
```
pascalchevrel: s/so/to

pascalchevrel:
There is a small typo but apart from this, I think this is fine to merge

Issue #610, opened 2016-02-07
Use Json class in Project class


pascalchevrel:
We have a Json class to get remote content, we should use it here:
https://github.com/mozfr/transvision/blob/master/app/classes/Transvision/Project.php#L81

PR #615, opened 2016-02-12
Fix issue #610

MatonAnthony:
Use the mozillal10n JSON class instead of json_encode()
Fix issue #610

### DIFF
--- a/app/classes/Transvision/Project.php
+++ b/app/classes/Transvision/Project.php
@@ -1,6 +1,8 @@
 <?php
 namespace Transvision;
 
+use Json\Json;
+
 /**
  * Project class
  *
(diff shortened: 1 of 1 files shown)


app/classes/Transvision/Project.php
```diff
@@ -1,5 +1,6 @@
 <?php
 namespace Transvision;
+use Zend\Json\Json;
```
pascalchevrel: Why Zend ? :)

app/classes/Transvision/Project.php
```diff
@@ -78,7 +80,8 @@ public static function getSupportedRepositories()
         // Read list of repositories from supported_repositories.json
         $file_name = APP_SOURCES . 'supported_repositories.json';
         if (file_exists($file_name)) {
-            $json_repositories = json_decode(file_get_contents($file_name), true);
+            $json_repositories = new Json($file_name);
+            $json_repositories = $json_repositories->fetchContent();
```
pascalchevrel: You can do that in one line which remains easy to read:
```php
$json_repositories = (new Json($file_name))->fetchContent();
```

MatonAnthony:
Fixed

Issue #594, opened 2016-01-19
add a  grep() method to the search class


pascalchevrel:
Small note to add a utility grep() function to our search class. Because almost every time we prepare a regex it is to grep data into it so we end up with code such as:

```php
$filtered_strings = preg_grep($search->getRegex(), $source_strings);
 ```
And this would be shorter and more readable IMO:
```php
$filtered_strings = $search->grep($source_strings);
```

pascalchevrel:
We are having an l10n hackathon mid-February and we will have a table with at least 4 people working on Transvision. I am putting that bug in the backlog of bugs to fix for the event.

PR #618, opened 2016-02-12
Fix issue #594

MatonAnthony:
Fix issue #594

Addition of a grep() method to search class to make the code easier to read.
Replacement of every occurence of `preg_grep($search->getRegex(), ` by `$search->grep(`

### DIFF
--- a/app/classes/Transvision/Search.php
+++ b/app/classes/Transvision/Search.php
@@ -225,4 +225,15 @@ class Search
     {
         return $this->regex_whole_words;
     }
+
+    /**
+     * Grep data in regex
+     *
+     * @param  array $source_strings The array of strings to be filtered
+     * @return array Return an array of filtered strings
+     */
+    public function grep($source_strings)
+    {
+        return preg_grep($this->regex, $source_strings);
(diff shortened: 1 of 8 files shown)


app/classes/Transvision/Search.php
```diff
@@ -225,4 +225,15 @@ public function isWholeWords()
...
         return $this->regex_whole_words;
     }
+
+    /**
+     * Grep data in regex
+     *
+     * @param  string $source_strings The string to be filtered
+     * @return string Return a filtered string
+     */
+    public function grep($source_strings)
```
pascalchevrel: Needs a unit test :)

app/classes/Transvision/Search.php
```diff
@@ -225,4 +225,15 @@ public function isWholeWords()
     {
         return $this->regex_whole_words;
     }
+
+    /**
+     * Grep data in regex
+     *
+     * @param  string $source_strings The string to be filtered
```
pascalchevrel: This is not a string, it's an array of strings

Issue #576, opened 2016-01-04
New suggestion API doesn't work without max_results parameter


pascalchevrel:
The new suggestion API should return results if the max_results option is not set, currently it returns an empty array, it probably just needs to get a default value for max_results.

PR #619, opened 2016-02-12
Fix issue #576 Suggestion API not working without max_results

MatonAnthony:
The issue was that the call was passing 0 as a parameter.

### DIFF
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -58,6 +58,9 @@ class ShowResults
         ];
         $flat_output = [];
 
+        // Reset $max_results as 10 if it's currently set to 0 (default value when the parameter is not specified in the API request).
+        $max_results = $max_results > 0 ? $max_results : 10;
+
         // Assign quality to each string in each group (source, target)
         foreach ($data as $group => $group_strings) {
             foreach ($group_strings as $single_string) {
--- a/tests/functional/api.php
+++ b/tests/functional/api.php
@@ -16,6 +16,8 @@ $paths = [
     ['v1/repositories/fr/', 200, '["aurora","beta","central","firefox_ios","gaia","gaia_2_0","gaia_2_1","gaia_2_2","gaia_2_5","mozilla_org","release"]'],
     ['v1/suggestions/central/en-US/fr/ar/?max_results=2', 200, '["Bookmark","Marque-page"]'],
     ['v1/suggestions/central/en-US/fr/ar/?max_results=10', 200, '["Bookmark","Bookmarks","New Bookmarks","Bookmark This Page","Marque-page","Marque-pages","Marquer cette page","Nouveaux marque-pages"]'],
+    ['v1/suggestions/central/en-US/fr/ar/?max_results=0', 200, '["Bookmark","Bookmarks","New Bookmarks","Bookmark This Page","Marque-page","Marque-pages","Marquer cette page","Nouveaux marque-pages"]'],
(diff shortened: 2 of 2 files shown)


app/models/api/suggestions.php
```diff
@@ -56,6 +56,14 @@
     return $value;
 };
 
+if (empty($get_option('max_results')) || $get_option('max_results') == 0) {
```
flodolo: The right approach is to evaluate `$max_results` inside the called method, not duplicate the call. The function is also very similar to `getTranslationMemoryResults`, so you can take inspiration for the check.

Also note that `$get_option('max_results')` can't really be empty.

MatonAnthony:
I think, it's good this time.

app/classes/Transvision/ShowResults.php
```diff
@@ -58,6 +58,8 @@ public static function getSuggestionsResults($source_strings, $target_strings, $
         ];
         $flat_output = [];
 
+        $max_results = $max_results <= 0 ? 10 : $max_results;
```
flodolo: Approach is definitely correct. Personally I would swap the check to make it more 'natural' (we want to consider the value provided only if its bigger than 0)
```PHP
$max_results = $max_results > 0 ? $max_results : 10;
```

app/classes/Transvision/ShowResults.php
```diff
@@ -58,6 +58,8 @@ public static function getSuggestionsResults($source_strings, $target_strings, $
         ];
         $flat_output = [];
 
+        $max_results = $max_results > 0 ? $max_results : 10;
```
pascalchevrel: I think it would be good to add a comment explaining the issue and why we have to check that $max_results isn't equal to 0.

It would be good to also add a functional test and I think we don't have unit tests for this method.

flodolo:
I think Pascal's request was to create a Unit test (which I didn't add when creating the function, so it's on me). Also, not sure about the functional test added, since we only have 8 results.

app/classes/Transvision/ShowResults.php
```diff
@@ -58,6 +58,9 @@ public static function getSuggestionsResults($source_strings, $target_strings, $
         ];
         $flat_output = [];
 
+        // Make sure $max_results is never set to 0, in which case we set it to 10
```
flodolo: I would rephrase this a bit: "Reset $max_results as 10 if it's currently set to 0 (default value when the parameter is not specified in the API request)."

MatonAnthony: @flodolo I don't have any idea about what i should be doing for the functional test, it was Pascal request so i did it.

flodolo:
Not sure what happened with all the changes...

MatonAnthony:
We found the issue, it'll be corrected in 5 minutes ;).
Php-cs-fixer went crazy with his config file

flodolo:
I don't think you have any way to fix, besides using master's files and adding back the changes you actually need.

As for the functional tests, I guess it's OK to leave them like this (the fix would be to add more strings), but I would still add some basic unit tests.

MatonAnthony:
Hello, I'll try to fix the first issues tomorrow and to find the time to
write the Unit test
Le 14 févr. 2016 19:55, "Francesco Lodolo" <notifications@github.com> a
écrit :

flodolo:
Note: instead of copying files, you can simply copy the file (or pieces of code) that you want to keep, reset you branch to master, then commit and force push.

pascalchevrel:
Merging, I will do a follow up patch to add a unit test, thanks.

PR #620, opened 2016-02-13
Adding prestissimo as a dev dependency

MatonAnthony:
Do not merge.
It's just to have the travis run to compare performance.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,7 +67,14 @@
     ],
 
     "config": {
-        "component-dir": "web/assets"
+        "component-dir": "web/assets",
+        "prestissimo": {
+            "maxConnections": 12,
+            "minConnections": 3,
+            "pipeline": true,
+            "verbose": true,
+            "insecure": false
+        }
     },
 
     "require": {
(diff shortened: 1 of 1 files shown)


MatonAnthony:
Travis build took 134 sec with prestissimo.

MatonAnthony:
Travis took 121 secon that build : https://travis-ci.org/mozfr/transvision/jobs/109031892
Using the default suggested configuration.

MatonAnthony:
It took 108 seconds with the last build : https://travis-ci.org/mozfr/transvision/jobs/109032440
I suggest we keep that. It implies to ask Travis to make 12 parallels request.

MatonAnthony:
@pascalchevrel What do you think about it ?

pascalchevrel:
I think it's fine but the dependency should be in the require-dev section of composer.json, not the require one, because we want to do that to speed up our dev environment but don't want to install that dependency on production.

MatonAnthony:
I think it's good this time.

pascalchevrel:
I am stull seeing it in the require section and not require-dev ;)

MatonAnthony:
Are you sure, you unfold the code ?
That what i'm seeing : 
capture d ecran 2016-02-19 a 14 52 01

pascalchevrel:
you're right, merging, thanks! :+1:

PR #623, opened 2016-02-13
Fix #604, adding two links for an API always working

### LINKED ISSUE #604: Some API calls don't return results while the main search does
From this results page [1], click on the API request link [2], the returned JSON is empty

[1] https://transvision.mozfr.org/?recherche=%C3%89cran&repo=aurora&sourcelocale=en-US&locale=fr&search_type=strings&case_sensitive=case_sensitive
[2] https://transvision.mozfr.org/?recherche=%C3%89cran&repo=aurora&sourcelocale=en-US&locale=fr&search_type=strings&case_sensitive=case_sensitive&json

Thegennok:
In the case of no result, the API result show no result, but the view show more result because of the other local. So whe add link for the two local in the API link. See #604.

### DIFF
--- a/app/classes/Transvision/Utils.php
+++ b/app/classes/Transvision/Utils.php
@@ -438,10 +438,31 @@ class Utils
      * This is used on views which also exist in our public API
      * https://github.com/mozfr/transvision/wiki/JSON-API
      *
-     * @return string URL with 'json' appended as part of the query string
+     * @param  boolean $revert if true, invert the locale and the source_locale in the API generated link
+     * @return string  URL with 'json' appended as part of the query string
      */
-    public static function redirectToAPI()
+    public static function redirectToAPI($revert = false)
     {
-        return $_SERVER["REQUEST_URI"] . (is_null($_SERVER['QUERY_STRING']) ? '?json' : '&json');
+        if (! is_null($_SERVER['QUERY_STRING'])) {
+            // Allow to revert the locale and the sourcelocale when redirecting to the API
+            if ($revert) {
+                $arg = [];
+                parse_str($_SERVER['QUERY_STRING'], $arg);
+                $sourcelocale = $arg['sourcelocale'];
+                $arg['sourcelocale'] = $arg['locale'];
+                $arg['locale'] = $sourcelocale;
+                $query = http_build_query($arg);
+            } else {
+                $query = $_SERVER['QUERY_STRING'];
+            }
+            $address = (strstr($_SERVER['REQUEST_URI'], '?') ?
+                strstr($_SERVER['REQUEST_URI'], '?', true) :
+                $_SERVER['REQUEST_URI']) . '?' .  $query;
+        } else {
+            $query = null;
+            $address = $_SERVER['REQUEST_URI'];
+        }
+
+        return $address . (is_null($query) ? '?json' : '&json');
     }
 }
--- a/app/views/templates/api_promotion.php
+++ b/app/views/templates/api_promotion.php
@@ -1,4 +1,19 @@
+<?php
+    if ($controller == 'mainsearch') {
+        ?>
+<p class="api_link">
+    <span>API</span>These results are also available as an API request for <a href="<?=\Transvision\Utils::redirectToAPI()?>"><?=$requested_sourcelocale?></a> or <a href="<?=\Transvision\Utils::redirectToAPI(true)?>"><?=$requested_locale?></a>.<br>
+    <a href="https://github.com/mozfr/transvision/wiki/JSON-API">Learn more about the Transvision API</a>.
+</p>
+<?php
+
+    } else {
+        ?>
 <p class="api_link">
     <span>API</span>These results are also available as an <a href="<?=\Transvision\Utils::redirectToAPI()?>">API request</a>.<br>
-    <a href="https://github.com/mozfr/transvision/wiki/JSON-API">Learn more about Transvision API</a>.
+    <a href="https://github.com/mozfr/transvision/wiki/JSON-API">Learn more about the Transvision API</a>.
 </p>
+<?php
+
+    }
+?>
--- a/tests/units/Transvision/Utils.php
+++ b/tests/units/Transvision/Utils.php
@@ -458,11 +458,13 @@ class Utils extends atoum\test
     {
         return [
             [
-                'http://transvision.mozfr.org/string/?entity=browser/chrome/browser/downloads/downloads.properties:stateStarting&repo=central',
-                'http://transvision.mozfr.org/string/?entity=browser/chrome/browser/downloads/downloads.properties:stateStarting&repo=central&json',
+                'http://transvision.mozfr.org/string/?recherche=screen&repo=aurora&sourcelocale=en-US&locale=fr&search_type=strings',
+                'http://transvision.mozfr.org/string/?recherche=screen&repo=aurora&sourcelocale=en-US&locale=fr&search_type=strings&json',
+                'http://transvision.mozfr.org/string/?recherche=screen&repo=aurora&sourcelocale=fr&locale=en-US&search_type=strings&json',
             ],
             [
                 'http://transvision.mozfr.org/api/v1/versions/',
+                'http://transvision.mozfr.org/api/v1/versions/?json',
                 'http://transvision.mozfr.org/api/v1/versions/?json',
             ],
         ];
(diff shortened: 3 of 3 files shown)


app/classes/Transvision/Utils.php
```diff
@@ -438,10 +438,32 @@ public static function ago($datetime, $ref_time = '')
...
-        return $_SERVER["REQUEST_URI"] . (is_null($_SERVER['QUERY_STRING']) ? '?json' : '&json');
+        if (! is_null($_SERVER['QUERY_STRING'])) {
+            if ($revert == true) {
+                $arg = [];
+                parse_str($_SERVER['QUERY_STRING'], $arg);
+                $source_locale = $arg["sourcelocale"];
+                $locale = $arg["locale"];
+                $arg["sourcelocale"] = $locale;
+                $arg["locale"] = $source_locale;
+                $query = http_build_query($arg);
```
flodolo: I'd wait for @pascalchevrel to see his opinion on the general approach. Having said that, this is clearer but we can swap the values with just one temp variable (also add a comment?).

app/classes/Transvision/Utils.php
```diff
@@ -438,10 +438,32 @@ public static function ago($datetime, $ref_time = '')
...
+            } else {
+                $query = $_SERVER['QUERY_STRING'];
+            }
+//            dump($_SERVER["REQUEST_URI"]);
+            $address = (strstr($_SERVER["REQUEST_URI"], '?') ?
+                        strstr($_SERVER["REQUEST_URI"], '?', true) :
```
flodolo: Drop the commented var dump above, also I think we settled to use only tab for the following lines in [these structures](https://github.com/mozfr/transvision/blob/master/app/classes/Transvision/API.php#L74-L76)

app/views/templates/api_promotion.php
```diff
@@ -1,4 +1,19 @@
+<?php
+    if ($controller == "mainsearch") {
+        ?>
+<p class="api_link">
+    <span>API</span>These results are also available as an API request for the <a href="<?=\Transvision\Utils::redirectToAPI()?>"><?=$requested_sourcelocale?> locale</a> or the <a href="<?=\Transvision\Utils::redirectToAPI(true)?>"><?=$requested_locale?> locale</a>.<br>
```
flodolo: I would go with a simpler '...API request for XX or YY'

Thegennok:
Updated, thanks for the reviews ;)

app/views/templates/api_promotion.php
```diff
@@ -1,4 +1,19 @@
+<?php
+    if ($controller == "mainsearch") {
```
TheoChevalier: single quotes should be enough here

app/classes/Transvision/Utils.php
```diff
@@ -438,10 +438,31 @@ public static function ago($datetime, $ref_time = '')
...
+     * @param  boolean $revert if true, invert the locale and the source_locale in the API generated link
+     * @return string  URL with 'json' appended as part of the query string
      */
-    public static function redirectToAPI()
+    public static function redirectToAPI($revert = false)
     {
-        return $_SERVER["REQUEST_URI"] . (is_null($_SERVER['QUERY_STRING']) ? '?json' : '&json');
+        if (! is_null($_SERVER['QUERY_STRING'])) {
+            // Allow to revert the locale and the sourcelocale when redirecting to the API
+            if ($revert == true) {
```
TheoChevalier: nit for consistency, if ($revert)

tests/units/Transvision/Utils.php
```diff
@@ -469,15 +471,19 @@ public function redirectToAPIDP()
...
      */
-    public function testRedirectToAPI($a, $b)
+    public function testRedirectToAPI($a, $b, $c)
     {
         $obj = new _Utils();
         $_SERVER["REQUEST_URI"] = $a;
         $_SERVER['QUERY_STRING'] = isset(parse_url($a)['query'])
             ? parse_url($a)['query']
             : null;
+        $_SERVER['HTTP_HOST'] = "http://" . parse_url($a)['host'];
```
TheoChevalier: nit single quotes here as well

app/classes/Transvision/Utils.php
```diff
@@ -438,10 +438,31 @@ public static function ago($datetime, $ref_time = '')
      * This is used on views which also exist in our public API
      * https://github.com/mozfr/transvision/wiki/JSON-API
      *
-     * @return string URL with 'json' appended as part of the query string
+     * @param  boolean $revert if true, invert the locale and the source_locale in the API generated link
+     * @return string  URL with 'json' appended as part of the query string
      */
-    public static function redirectToAPI()
+    public static function redirectToAPI($revert = false)
```
pascalchevrel: I think $swap_locales would be more explicit

app/classes/Transvision/Utils.php
```diff
@@ -438,10 +438,31 @@ public static function ago($datetime, $ref_time = '')
...
     {
-        return $_SERVER["REQUEST_URI"] . (is_null($_SERVER['QUERY_STRING']) ? '?json' : '&json');
+        if (! is_null($_SERVER['QUERY_STRING'])) {
+            // Allow to revert the locale and the sourcelocale when redirecting to the API
+            if ($revert) {
+                $arg = [];
```
pascalchevrel: the array is created by parse_str(), you don't need to initialize it.

app/classes/Transvision/Utils.php
```diff
@@ -438,10 +438,31 @@ public static function ago($datetime, $ref_time = '')
...
+                parse_str($_SERVER['QUERY_STRING'], $arg);
```
pascalchevrel: I would preger $args to $arg because we have multiple arguments

app/classes/Transvision/Utils.php
```diff
@@ -438,10 +438,31 @@ public static function ago($datetime, $ref_time = '')
...
+                $sourcelocale = $arg['sourcelocale'];
+                $arg['sourcelocale'] = $arg['locale'];
+                $arg['locale'] = $sourcelocale;
+                $query = http_build_query($arg);
```
pascalchevrel: That whole part for swapping locales is ok but I don't think you should build the query at that time

app/classes/Transvision/Utils.php
```diff
@@ -438,10 +438,31 @@ public static function ago($datetime, $ref_time = '')
...
+            } else {
+                $query = $_SERVER['QUERY_STRING'];
+            }
+            $address = (strstr($_SERVER['REQUEST_URI'], '?') ?
+                strstr($_SERVER['REQUEST_URI'], '?', true) :
+                $_SERVER['REQUEST_URI']) . '?' .  $query;
```
pascalchevrel: That ternary has a strange syntax, the parenthesis seem to be mixed.

tests/units/Transvision/Utils.php
```diff
@@ -458,12 +458,14 @@ public function redirectToAPIDP()
     {
         return [
             [
-                'http://transvision.mozfr.org/string/?entity=browser/chrome/browser/downloads/downloads.properties:stateStarting&repo=central',
-                'http://transvision.mozfr.org/string/?entity=browser/chrome/browser/downloads/downloads.properties:stateStarting&repo=central&json',
+                'http://transvision.mozfr.org/string/?recherche=screen&repo=aurora&sourcelocale=en-US&locale=fr&search_type=strings',
+                'http://transvision.mozfr.org/string/?recherche=screen&repo=aurora&sourcelocale=en-US&locale=fr&search_type=strings&json',
+                'http://transvision.mozfr.org/string/?recherche=screen&repo=aurora&sourcelocale=fr&locale=en-US&search_type=strings&json',
```
pascalchevrel: Those urls wouldn't work in Transvision, even though this is a syntethic test, I think you should use real URIs

tests/units/Transvision/Utils.php
```diff
@@ -471,15 +473,19 @@ public function redirectToAPIDP()
...
-    public function testRedirectToAPI($a, $b)
+    public function testRedirectToAPI($a, $b, $c)
     {
         $obj = new _Utils();
-        $_SERVER["REQUEST_URI"] = $a;
+        $_SERVER['REQUEST_URI'] = $a;
         $_SERVER['QUERY_STRING'] = isset(parse_url($a)['query'])
             ? parse_url($a)['query']
             : null;
+        $_SERVER['HTTP_HOST'] = 'http://' . parse_url($a)['host'];
```
pascalchevrel: I don't think you should do that because you are adding even more transformation logic in your tests of globals, so you are testing something that is no longer exactly what the real function returns. Instead of rebuilding a full URL to match what we have in the data provider function, I think you should just remove the host from the strings in the data provider and only keep it as a set of REQUEST_URIs because this is what redirectToAPI() works from and returns.

tests/units/Transvision/Utils.php
```diff
@@ -458,12 +458,14 @@ public function redirectToAPIDP()
     {
         return [
             [
-                'http://transvision.mozfr.org/string/?entity=browser/chrome/browser/downloads/downloads.properties:stateStarting&repo=central',
-                'http://transvision.mozfr.org/string/?entity=browser/chrome/browser/downloads/downloads.properties:stateStarting&repo=central&json',
```
pascalchevrel: With the removal of these tests that contained slashes in entity names, it actually hides the fact that your new version of the function returns escaped slashes instead of real slashes.

pascalchevrel:
merging, I will do a follow up clean up patch

Issue #593, opened 2016-01-19
In debug mode, send perf metrics in custom http header


pascalchevrel:
Currently we can log our metrics in debug mode in the console (or the apache logs if one does not use the php dev server).

I think it would be convenient to also send this data in a custom  http header when in debug mode, this way we can see on the beta server or our VPSs wat the perf of a page is without having to ssh and look at the logs. HTTP headers are also more convenient than outputing directly this infoirmation in the view because the views we need most performance from are the json api and we can't ourput data there.

I quickly checked and it's easy to do, see in the screenshot below the result in firebug, I called the .header Transvision-perf.

![entetes](https://cloud.githubusercontent.com/assets/454375/12425795/c89cc22e-bed6-11e5-8ac0-fbc7b0779c3d.png)

I have something tthat works now but maybe I should let it to a new contributor? Do you like the idea?

pascalchevrel:
We are having an l10n hackathon mid-February and we will have a table with at least 4 people working on Transvision. I am putting that bug in the backlog of bugs to fix for the event.

flodolo:
(side note: the header is visible also in Vanilla devtools: network tab, click on the 200 response, right column)

PR #624, opened 2016-02-14
Add HTTP debug header, #593

Thegennok:
As requested in #593, there is now the memory comsuption and the time elapsed in a HTTP header if perf_debug is true

### DIFF
--- a/app/classes/Transvision/Utils.php
+++ b/app/classes/Transvision/Utils.php
@@ -360,6 +360,7 @@ class Utils
                            . round((microtime(true) - $_SERVER['REQUEST_TIME_FLOAT']), 4);
             error_log($memory);
             error_log($render_time);
+            header('Transvision-perf: ' . $memory . '; ' . $render_time);
         }
     }
 


flodolo:
Since the code is basically the same of `logScriptPerformances`, I think it makes more sense to only keep one function and pass a parameter to ask to log info in error_log, header, or both

app/classes/Transvision/Utils.php
```diff
@@ -364,6 +364,26 @@ public static function logScriptPerformances()
...
+    {
+        if (DEBUG && PERF_CHECK) {
+            $memory = 'Memory peak '
+                      . memory_get_peak_usage(true)
+                      . ' ('
+                      . round((memory_get_peak_usage(true) / (1024 * 1024)), 2)
+                      . 'MB)';
+            $render_time = 'Elapsed time (s) '
+                           . round((microtime(true) - $_SERVER['REQUEST_TIME_FLOAT']), 4);
+            header('Transvision-perf: ' . $memory . ' ; ' . $render_time);
```
flodolo: No need for an extra space before the semicolon.
nit for the description: http header -> HTTP header

Thegennok:
He log in error_log and HTTP header in the same time now.

flodolo:
Unfortunately this generates an error in the log:
```
PHP Warning:  Cannot modify header information - headers already se
nt by (output started at /home/flodolo/github/transvision/app/views/templates/base.php:162) in /home/flodolo/github/transvision/app/classes/Transvision/
Utils.php on line 363, referer: xxx
```

pascalchevrel:
This is too simplistic an approch :) 
We need all views to output that perf http header and we need to send that header before the real content otherwise we get an error about headers already sent.

The right approach is to split the function in 3:
getScriptPerformances() that returns an array of raw data
logScriptPerformances() that does what it does now but gets the data from getScriptPerformances()
outputPerformancesHHeader() that outputs the header and also gets the data from getScriptPerformances()

Then in the dispatcher, we use output buffering and at the end of the dispatcher, we print the buffers like that:
```php
print $perf_header . $content;
```
That's actually moving what we do in base.php to the dispatcher.
The json view should also be updated to not die and only print.

I have a working patch so maybe I could take the bug.

flodolo:
Replaced by #637

Issue #577, opened 2016-01-04
Use woff2 fonts by default


TheoChevalier:
Fonts download is an important part of our page load time:

![capture du 2016-01-04 17-34-07](https://cloud.githubusercontent.com/assets/1294206/12094968/0cfec22c-b30c-11e5-88a7-1bb9f469d7a9.png)

woff2 fonts are usually smaller files than woff, so we could use this tool https://everythingfonts.com/woff-to-woff2 to convert our current files and propose woff2 first to compatible browsers: http://caniuse.com/#feat=woff2 (Since Firefox is compatible, most of our visitors should benefit from this)

TheoChevalier:
Fixed by #625 :)

PR #625, opened 2016-02-14
Issue #577 – Transition from woff to woff2 fonts

clement-lefevre:
In issue #577 it was required to move from woff to woff2 for fonts files.

It was suggested to convert the files, however I found some downloadable versions that proved to be lighter than converted versions and then checked on specific locales like ro, fu and so on and saw everything.

So it seems that those one are corrects. I can take care of providing converted versions at any time if some error are discovered at any time.


TheoChevalier:
Please keep woff fonts for compatibility with browsers not yet supporting woff2 
Example: https://github.com/mozfr/nightly/blob/master/web/assets/style/base.css#L2

flodolo:
(which is [a lot of browsers](http://caniuse.com/#feat=woff2)...

TheoChevalier:
(both woff and woff2 files are missing in the commit)

TheoChevalier:
wondering if woff2 only loads with HTTPS

clement-lefevre:
Seems better when src: is not overwritten and redefined.
woff2 fonts are now correctly downloaded.

TheoChevalier:
lgtm now

PR #628, opened 2016-02-14
Fixes #603

### LINKED ISSUE #603: Can't search for strings containing encoded characters like \u0027 (apostrophe)
In some Android files for instance, we can't use apostrophes, so we're encoding them this way \u0027 in the translations.
This query should return results, but currently it doesn't: https://transvision.mozfr.org/?recherche=%C3%89cran+d\u0027accueil&repo=aurora&sourcelocale=en-US&locale=fr&search_type=strings

Another place where we're using such characters: https://transvision.mozfr.org/?recherche=remotetabs-panelmenu.label&repo=aurora&sourcelocale=en-US&locale=fr&search_type=entities#browser_chrome_browser_customizableui_customizableWidgets.properties_remotetabs-panelmenu.label
\u00AD is a special character that prevents hyphenation.

The Transvision bug is that the backslash character doesn't get escaped, this query works (notice the "\\\"):

https://transvision.mozfr.org/?recherche=Onglets+\\u00ADsynchronis%C3%A9s&repo=aurora&sourcelocale=en-US&locale=fr&search_type=strings

benoit-l:
* Removes an unneeded call to stripslashes
* Removes a test that didn't actually test anything except for the PHP parser itself
* Test that backslashes are not removed again when escaping unicode entities (even if they possibly need to be in other cases)

### DIFF
--- a/app/classes/Transvision/Utils.php
+++ b/app/classes/Transvision/Utils.php
@@ -276,9 +276,6 @@ class Utils
             return '';
         }
 
-        // Remove escaped characters (quotes)
-        $string = stripslashes($string);
-
         // Filter out double spaces
         $string = Strings::mtrim($string);
 
--- a/tests/units/Transvision/Utils.php
+++ b/tests/units/Transvision/Utils.php
@@ -205,8 +205,8 @@ class Utils extends atoum\test
                 'toto',
             ],
             [
-                'don\'t escape',
-                "don't escape",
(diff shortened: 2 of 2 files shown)


flodolo:
I can't really see anything wrong with the approach, also after testing it a bit locally.

Let's see if @pascalchevrel has thoughts before merging.

pascalchevrel:
LTM, thanks

PR #631, opened 2016-02-15
Channel comparison: Improve view for empty comparison

flodolo:


app/views/channelcomparison.php
```diff
@@ -30,21 +30,26 @@
...
-     "    <th colspan='3'>Locale: {$locale}</th>\n" .
-     "  </tr>\n" .
-     "  <tr>\n" .
-     "    <th>Key</th>\n" .
-     "    <th>{$chan1}</th>\n" .
-     "    <th>{$chan2}</th>\n" .
-     "  </tr>\n";
+if (empty($common_strings)) {
+    echo "<h3>Comparison is empty</h3>\n" .
+         "<p class='subtitle'>There are no strings to compare: this locale is not supported for both channels.</p>\n";
```
pascalchevrel: If the strings are identical for the locale on two repositories (beta and release for example), then the array will be empty, yet the locale is supported on both channels. Ex Acholi:
https://transvision.mozfr.org/channelcomparison/?locale=ach&chan1=beta&chan2=release

flodolo: Uhm, true that. Just: 'There are no strings to compare between the selected channels."?

pascalchevrel: or "There are no string differences for this locale between the A and B channels" (A and B being the repo namesà, because there are strings to compare, they are just identical.

app/views/channelcomparison.php
```diff
@@ -30,21 +30,26 @@
...
-     "    <th colspan='3'>Locale: {$locale}</th>\n" .
-     "  </tr>\n" .
-     "  <tr>\n" .
-     "    <th>Key</th>\n" .
-     "    <th>{$chan1}</th>\n" .
-     "    <th>{$chan2}</th>\n" .
-     "  </tr>\n";
+if (empty($common_strings)) {
+    echo "<h3>Comparison is empty</h3>\n" .
+         "<p class='subtitle'>There are no string differences for this locale between the {$repos_nice_names[$chan1]} and {$repos_nice_names[$chan2]} channels.</p>\n";
```
flodolo: Updated string following suggestion.

PR #632, opened 2016-02-18
Follow up on pull request #625 - some font files are not loading

TheoChevalier:
Didn't catch that during the review, but there were missing commas in 4a8e33bb56ce71a70ee70d9e3b7e64fe19291c19 preventing some fonts from loading.


flodolo:
Looks good :+1:

PR #635, opened 2016-02-22
Refactor glossaire.sh to create TMX only when necessary

### LINKED ISSUE #634: Improve speed of updates: build glossaries only when necessary
After looking at the update process for a while, I noticed a few things that could be improved, but I need to figure out if that's possible and how:

- [x] Avoid resetting the repo with `hg up -C`, we don't have local changes to purge.
That was fixed by #629 

- [ ] Creating TMX cache is not a fast process (I have 705 of those in my VM), and more important we create a lot of them needlessly. We should rebuild a cache only if:
* English changed. Which means comparing the old TMX with the new one, since English will basically have changes all the time.
* If English didn't change, build the locale only if `hg pull -u` actually pulled some changesets. [This](http://stackoverflow.com/questions/8922787/mercurial-check-whether-last-pull-update-introduced-changes) looks interesting.

flodolo:
I also took the chance to fix some style/coding issues:
- Don't use UPPERCASE for variable names, leave it only for system variables like `$PATH`.
- Don't check boolean variables with `if $boolean_var`, since it's [not necessarily safe](http://stackoverflow.com/questions/2953646/how-to-declare-and-use-boolean-variables-in-shell-script).

The final rewrite is quite heavy, but the idea is pretty simple:
- Instead of having a `createTMX` variable, we now have a `forceTMX`, which can be used to force TMX creation without pulling the repos (e.g. you manually change a string in a repository).
- When updating a locale's repository, we store information if the hg command actually pulled changesets. No point in recreating a TMX if there are no changes inside the locale.
– We always build the TMX for en-US, and check if it changes after a pull. All repositories have a lot of code changes, but that doesn't mean that there are new strings. If en-US has changed, we rebuild all locales, even if there were no changes in the l10n repository.

For mozilla.org and iOS we always create the TMX, I don't think it makes sense to complicate the code to threat them differently. I've also removed the L20n test, I don't think it makes much sense to keep it around at this point.

P.S. I tried to embed `buildCache` into `updateLocale`, but there are way too many variables to pass around to make it happen.

Fixes #634

### DIFF
--- a/app/scripts/glossaire.sh
+++ b/app/scripts/glossaire.sh
@@ -1,8 +1,8 @@
 #! /usr/bin/env bash
 
 # Syntax:
-# - without parameters: update all locales
-# - one parameter (locale code): update only the requested locale
+# - locale code: update only the requested locale
+# - 'no-snapshot': avoid creating a data snapshot
 
 function interrupt_code()
 # This code runs if user hits control-c
@@ -15,21 +15,21 @@ function interrupt_code()
 trap interrupt_code SIGINT
 
 # Pretty printing functions
-NORMAL=$(tput sgr0)
-GREEN=$(tput setaf 2; tput bold)
-RED=$(tput setaf 1)
+standard_color=$(tput sgr0)
+green=$(tput setaf 2; tput bold)
+red=$(tput setaf 1)
 
 function echored() {
-    echo -e "$RED$*$NORMAL"
+    echo -e "$red$*$standard_color"
 }
 
 function echogreen() {
-    echo -e "$GREEN$*$NORMAL"
+    echo -e "$green$*$standard_color"
 }
 
 function echo_manual() {
-    echo "ERROR: too many or incorrect arguments. "
-    echo "Run 'glossaire.sh' without parameters to update all locales"
+    echo "Run 'glossaire.sh' without parameters to update all locales."
+    echo "Run 'glossaire.sh help' to display this manual."
     echo "---"
     echo "To update only one locale, add the locale code as first parameter"
     echo "(e.g. 'glossaire.sh fr' to update only French)."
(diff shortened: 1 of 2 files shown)


flodolo:
The first test on my VPS is pretty encouraging: the entire setup+glossaire took less than 8 minutes, and updated a couple of locales on aurora as expected.

Obviously the time will be still ~30 minutes on merge day, but it will be an issue only once every 6 weeks.

app/scripts/glossaire.sh
```diff
@@ -15,16 +15,16 @@ function interrupt_code()
...
-NORMAL=$(tput sgr0)
-GREEN=$(tput setaf 2; tput bold)
-RED=$(tput setaf 1)
+standard_color=$(tput sgr0)
+green=$(tput setaf 2; tput bold)
+red=$(tput setaf 1)
 
 function echored() {
-    echo -e "$RED$*$NORMAL"
+    echo -e "$app_folder$*$standard_color"
```
pascalchevrel: $app_folder ?

flodolo: Argh, some poor copy and paste while fixing the variable cases :-\

flodolo:
Code updated (left separate commits ).

Logs on both my VM and VPS look fine (they're creating cache for locales that had updates, and they're going through Gaia).

pascalchevrel:
I didn't find any bug, the only thing I would suggest while we are at refactoring glossaire would be to add a help parameter:
```bash
    if [ "$1" == "help" ]
    then
        echo_manual
        exit 1
    elif [ "$1" == "no-snapshot" ]
    then
        create_snapshot=false
    else
        all_locales=false
        locale_code=$1
    fi
```

flodolo:
Added as a separate commit, let me know if I should squash them.

pascalchevrel:
This is fine, please squash the commits thanks :)

PR #637, opened 2016-02-22
Issue #593: add http headers with perf metrics

pascalchevrel:
- New header: Transvision-perf
- Does not depend on DEBUG of PERF_CHECK to be checked because we don't impact the server logs with http headers and it allows getting perf metrics on production easily

### DIFF
--- a/app/classes/Transvision/Utils.php
+++ b/app/classes/Transvision/Utils.php
@@ -340,24 +340,48 @@ class Utils
     }
 
     /**
-     * Utility function to log the memory used by a script
-     * and the time needed to generate the page
+     * Utility function to return the memory used by a script
+     * and the time needed to compute the data.
+     *
+     * @return array [Memory peak in bytes, Memory peak in MB, Computation time]
+     */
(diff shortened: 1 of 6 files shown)


app/classes/Transvision/Utils.php
```diff
@@ -340,27 +340,51 @@ public static function getOrSet($arr, $value, $fallback)
...
     /**
+     * Utility function to log the memory used by a script
+     * and the time needed to generate the page as an http header.
+     *
+     * @return void
+     */
+    public static function addPerformancesHTTPHeader()
+    {
+        list($memory_peak_B, $memory_peak_MB, $computation_time) = self::getScriptPerformances();
+        header("Transvision-perf: Memory: {$memory_peak_B} ({$memory_peak_MB}MB) ; Time: {$computation_time}s");
```
flodolo: nit: no space before semicolon, also http->HTTP in the function description.

flodolo:
Can't spot any issue, and it works great :+1:

PR #643, opened 2016-02-23
Add script to list/delete obsolete files

### LINKED ISSUE #636: iOS TMX script should ignore README.md
Currently tries to create a TMX for the README, and also create a folder in the TMX folder.

flodolo:
Fix #636 and also #636 (ignore README.md for iOS).

Run as `clean_data.py` without arguments to only have a list of the extra files, I'd be interested to know if you see any weird folder/file in your installations. 

Run as `clean_data.py --delete` to actually delete the folders, but you might encounter some permission problems depending on your setup.

I don't think it's something we should run automatically, but it's good to have and run periodically to double check if we need to clean up something.

For example I noticed (and [fixed](https://github.com/mozilla-l10n/mozilla-l10n-query/commit/3b666379385509101e936d83b43b966c0e208a54)) several locales missing for iOS and mozilla.org

### DIFF
--- a/app/scripts/clean_data.py
+++ b/app/scripts/clean_data.py
@@ -1,0 +1,155 @@
+#!/usr/bin/env python
+
+import argparse
+import json
+import glob
+import os
+import shutil
+import sys
+from ConfigParser import SafeConfigParser
+
+
+def main():
+    # Parse command line options
+    cl_parser = argparse.ArgumentParser()
+    cl_parser.add_argument('--delete',
+                           help='Delete files', action='store_true')
+    args = cl_parser.parse_args()
+
+    # Get absolute path of ../config from the current script location (not the
+    # current folder)
+    config_folder = os.path.abspath(
+        os.path.join(os.path.dirname(__file__), os.pardir, 'config'))
+
+    # Read Transvision's configuration file from ../config/config.ini
+    config_file = os.path.join(config_folder, 'config.ini')
+    if not os.path.isfile(config_file):
+        print 'Configuration file /app/config/config.ini is missing.'
+        sys.exit(1)
+    else:
+        config_parser = SafeConfigParser()
+        config_parser.read(config_file)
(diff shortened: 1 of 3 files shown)


pascalchevrel:
For the record here is the output for me locally:
Analyzing sources in config/sources
gaia_2_1.txt is not a supported repository.
l20n_test.txt is not a supported repository.
gaia_2_0.txt is not a supported repository.
gaia_2_2.txt is not a supported repository.

Analyze release
ur is not a supported locale

Analyze cache files in TMX
/home/pascalc/transvisiondata/TMX/de/cache_de_gaia_2.php is not a known cache file
/home/pascalc/transvisiondata/TMX/README.md/cache_README.md_firefox_ios.php is not a known cache file
/home/pascalc/transvisiondata/TMX/pl/cache_pl_l20n_test.php is not a known cache file
/home/pascalc/transvisiondata/TMX/fr/cache_fr_l20n_test.php is not a known cache file

flodolo:
Looks good. In my case the installation was very old, I found about 60/70 elements to remove.

app/scripts/clean_data.py
```diff
@@ -0,0 +1,153 @@
...
+                if args.delete:
+                    full_path = os.path.join(folder_path, folder)
+                    print "Removing folder:", full_path
+                    shutil.rmtree(full_path)
+    if not need_cleanup:
+        print "Nothing to remove."
+
+    # Check cache files
+    print '--\nAnalyze cache files in TMX'
+    excluded_folders = [
```
pascalchevrel: I think it would be valuable to put a comment here explaining why we exclude fur and hi

flodolo: Good point. We actually removed 'fur' the other day, the only one remaining and it should be fixed on the extraction script side. Added a commit for that.

pascalchevrel:
The rest of the patch looks fine to me, I think you can squash and merge

PR #648, opened 2016-02-25
Create view to display translation inconsistencies

### LINKED ISSUE #410: Create a view listing translation inconsistencies
Besnik would like a view that would list all the inconsistencies in translation that may happen when the same English string is used in several contexts but there are different translations for them. The differences in translation can be on purpose, but not always.

Ex:
English uses "Add to bookmark" in two different entities, but the locale uses two different translations, /ex in French 'Ajouter aux marque-pages' and 'Ajout aux marque-pages'.

The view should find all duplicate English strings for a repo and then check if the translations are different too. If there are different, present the data in a way that is understandable.

flodolo:
I think this should be ready for a review.

One note: I didn't use "perfect search" for the search link because it fails quite often, for example with HTML fragments coming from DTD with leading/trailing new lines.

Also added a "splash screen" to avoid burning resources (and let people wait for seconds) when randomly open the menu.

Fixes #410

### DIFF
--- a/app/classes/Transvision/Consistency.php
+++ b/app/classes/Transvision/Consistency.php
@@ -1,0 +1,117 @@
+<?php
+namespace Transvision;
+
+/**
+ * Consistency class
+ *
+ * Functions used to analyze translation consistency
+ *
+ * @package Transvision
+ */
+class Consistency
+{
+    /**
+     * Find duplicated strings (strings with the same text, ignoring case).
+     *
+     * @param array $strings_array Array of strings in the form
+     *                             string_id => string_value
+     *
+     * @return array Array of duplicated strings, same format as the
+     *               input array
+     */
+    public static function findDuplicates($strings_array)
+    {
+        $duplicates = [];
+        // Use array_filter to exclude empty strings
+        $strings_array = array_filter($strings_array);
+        asort($strings_array);
+
+        $previous_key = '';
+        $previous_value = '';
+        foreach ($strings_array as $key => $value) {
+            if (strcasecmp($previous_value, $value) === 0) {
+                $duplicates[$previous_key] = $previous_value;
+                $duplicates[$key] = $value;
+            }
+            $previous_value = $value;
+            $previous_key = $key;
+        }
+
+        return $duplicates;
+    }
+
+     /**
+      * Filter out strings that should not be evaluted for consistency.
+      *
+      * @param  array $strings_array Array of strings in the form
+      *                              string_id => string_value
+      * @param  string Repository identifier
+      *
+      * @return array Array of filtered strings, with known false positives
+      *               removed
+      */
+     public static function filterStrings($strings_array, $repo)
+     {
+         if (in_array($repo, Project::getDesktopRepositories())) {
+             $repository_type = 'desktop';
+         } elseif (in_array($repo, Project::getGaiaRepositories())) {
+             $repository_type = 'gaia';
+         } else {
+             $repository_type = $repo;
+         }
+
+         // Determine if a string should be excluded
+         $ignore_string = function ($key, $value) use ($repository_type) {
+             // Ignore strings of length 1 (e.g. accesskeys) and empty strings.
+             if (strlen($value) <= 1) {
+                 return true;
+             }
+
+             // Exclude CSS rules, "width:" or "height:"
+             if (Strings::inString($value, ['width:', 'height:'])) {
+                 return true;
+             }
+
+             // Exclude CSS width values like '38em'
+             if (preg_match('/[\d|\.]+em/', $value)) {
+                 return true;
+             }
+
+             if ($repository_type == 'gaia') {
+                 // Ignore plural forms containing [] in the key
+                 if (Strings::inString($key, ['[', ']'])) {
+                     return true;
+                 }
+
+                 // Ignore accessibility strings
+                 if (strpos($key, 'accessibility.properties') !== false) {
+                     return true;
+                 }
+             }
+
+             if ($repository_type == 'desktop') {
+                 /*
+                    Ignore some specific files:
+                    - AccessFu.properties: there are accessibility
+                      strings that remain identical for English in full and
+                      abbreviated form.
+                    - defines.inc (language pack attribution)
+                    - region.properties
+                 */
+                 if (Strings::inString($key, ['AccessFu.properties', 'defines.inc', 'region.properties'])) {
+                     return true;
+                 }
+             }
+
+             return false;
+         };
+
+         foreach ($strings_array as $key => $value) {
+             if ($ignore_string($key, $value)) {
+                 unset($strings_array[$key]);
+             }
+         }
+
+         return $strings_array;
+     }
+}
--- a/app/classes/Transvision/Strings.php
+++ b/app/classes/Transvision/Strings.php
@@ -64,15 +64,29 @@ class Strings
     }
 
     /**
-     * Check if $needle is in $haystack
+     * Check if $needles are in $haystack
      *
-     * @param  string  $haystack String to analyse
-     * @param  string  $needle   The string to look for
-     * @return boolean True if the $haystack string contains $needle
+     * @param string  $haystack  String to analyze
+     * @param mixed   $needles   The string (or array of strings) to look for
+     * @param boolean $match_all True if we need to match all $needles, false
+     *                           if it's enough to match one. Default: false
+     *
+     * @return boolean True if the $haystack string contains any/all $needles
      */
-    public static function inString($haystack, $needle)
+    public static function inString($haystack, $needles, $match_all = false)
     {
-        return mb_strpos($haystack, $needle, $offset = 0, 'UTF-8') !== false ? true : false;
+        $matches = 0;
+        foreach ((array) $needles as $needle) {
+            if (mb_strpos($haystack, $needle, $offset = 0, 'UTF-8') !== false) {
+                // If I need to match any needle, I can stop at the first match
+                if (! $match_all) {
+                    return true;
+                }
+                $matches++;
+            }
+        }
+
+        return $matches == count($needles);
     }
 
     /**
--- a/app/models/consistency.php
+++ b/app/models/consistency.php
@@ -1,0 +1,75 @@
+<?php
+namespace Transvision;
+
+// Build arrays for the search form, ignore mozilla.org
+$channel_selector = Utils::getHtmlSelectOptions(
+    array_diff($repos_nice_names, ['mozilla.org']),
+    $repo,
+    true
+);
+
+$target_locales_list = Utils::getHtmlSelectOptions(
+    Project::getRepositoryLocales($repo),
+    $locale
+);
+
+$reference_locale = Project::getReferenceLocale($repo);
+
+// Set a default for the number of strings to display
+$strings_number = 0;
+
+$source_strings = Utils::getRepoStrings($reference_locale, $repo);
+// Ignore empty strings in translations
+$target_strings = array_filter(Utils::getRepoStrings($locale, $repo));
+
+if (! empty($source_strings)) {
+    // Remove known problematic strings
+    $duplicated_strings_english = Consistency::filterStrings($source_strings, $repo);
+    // Find strings that are identical in English
+    $duplicated_strings_english = Consistency::findDuplicates($duplicated_strings_english);
+}
+
+if (! empty($duplicated_strings_english)) {
+    $inconsistent_translations = [];
+    $available_translations = [];
+
+    /*
+        Using CachingIterator to be able to look ahead during the foreach
+        cycle, since it's an associative array.
+        http://php.net/manual/en/class.cachingiterator.php
+    */
+    $collection = new \CachingIterator(new \ArrayIterator($duplicated_strings_english));
+    foreach ($collection as $key => $value) {
+        // Ignore this string if is not available in the localization
+        if (! isset($target_strings[$key])) {
+            $available_translations = [];
+            continue;
+        }
+
+        $available_translations[] = $target_strings[$key];
+        /*
+            If the current English string is different from the previous one,
+            or I am at the last element, I need to store it with all available
+            translations collected so far.
+
+            $collection->getInnerIterator()->current() stores the value of
+            the next element.
+        */
+        if (! $collection->hasNext() || $collection->getInnerIterator()->current() != $value) {
+            $available_translations = array_unique($available_translations);
+            /*
+                Store this string only if there's more than one translation.
+                If there's only one, translations are consistent.
+            */
+            if (count($available_translations) > 1) {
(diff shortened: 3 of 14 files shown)


app/classes/Transvision/Consistency.php
```diff
@@ -0,0 +1,121 @@
+<?php
+namespace Transvision;
+
+/**
+ * Consistency class
+ *
+ * Functions used to analyze consistency
```
pascalchevrel: I would precise "to analyse translation consistency".

app/classes/Transvision/Consistency.php
```diff
@@ -0,0 +1,121 @@
...
+         } elseif (in_array($repo, Project::getGaiaRepositories())) {
+             $repository_type = 'gaia';
+         } else {
+             $repository_type = $repo;
+         }
+
+         $ignore_string = function ($key, $value) use ($repository_type) {
+             // Determine if a string should be excluded.
+
+             // Ignore strings of lenght 1 (e.g. accesskeys) and empty strings.
```
pascalchevrel: length

app/classes/Transvision/Consistency.php
```diff
@@ -0,0 +1,121 @@
...
+         if (in_array($repo, Project::getDesktopRepositories())) {
+             $repository_type = 'desktop';
+         } elseif (in_array($repo, Project::getGaiaRepositories())) {
+             $repository_type = 'gaia';
+         } else {
+             $repository_type = $repo;
+         }
+
+         $ignore_string = function ($key, $value) use ($repository_type) {
+             // Determine if a string should be excluded.
```
pascalchevrel: put the comment before the anonymous function to make it readable with folding on

app/classes/Transvision/Consistency.php
```diff
@@ -0,0 +1,121 @@
...
+
+             // Ignore strings of lenght 1 (e.g. accesskeys) and empty strings.
+             if (strlen($value) <= 1) {
+                 return true;
+             }
+
+             // Exclude CSS rules, "widht:" or "height:"
+             if (strpos($value, 'width:') !== false || strpos($value, 'height:') !== false) {
```
pascalchevrel: We have a Strings::inString() method, I think it would be interesting to extend it to support arrays like that:
```php
/**
 * Check if $needles are in $haystack
 *
 * @param  string  $haystack String to analyse
 * @param  array   $needles  The strings to look for
 * @return boolean True if the $haystack string contains all strings in $needles
 */
public static function inString($haystack, $needles)
{
    foreach ((array) $needles as $needle) {
        if (mb_strpos($haystack, $needle, $offset = 0, 'UTF-8') === false) {
            return false;
        }
    }

    return true;
}
```
Then you could replace the above check with:
```php
if (Strings::inString($value, ['width:', 'height:'])) {
    return true;
}
```
I could not measure any performance hit in this case because of the use of mb_strpos vs strpos

app/classes/Transvision/Consistency.php
```diff
@@ -0,0 +1,121 @@
...
+                 // Ignore plural forms containing [] in the key
+                 if (strpos($key, '[') !== false && strpos($key, '[') !== false) {
+                     return true;
+                 }
+
+                 // Ignore accessibility strings
+                 if (strpos($key, 'accessibility.properties') !== false) {
+                     return true;
+                 }
+             } elseif ($repository_type == 'desktop') {
```
pascalchevrel: you probably don't need an elseif and can just use an if

app/classes/Transvision/Consistency.php
```diff
@@ -0,0 +1,121 @@
...
+             }
+
+             // Exclude CSS widht values like '38em'
+             if (preg_match('/[\d|\.]+em/', $value)) {
+                 return true;
+             }
+
+             if ($repository_type == 'gaia') {
+                 // Ignore plural forms containing [] in the key
+                 if (strpos($key, '[') !== false && strpos($key, '[') !== false) {
```
pascalchevrel: Did you mean to search for [ and ]? I think there is a typo since you are looking for [ twice. Also, would be another candidate for `` if (Strings::inString($value, ['[', ']']))```

app/classes/Transvision/Consistency.php
```diff
@@ -0,0 +1,121 @@
...
+                 }
+
+                 /*
+                     Ignore AccessFu.properties, since there are accessibility
+                     strings that remain identical for English in full and
+                     abbreviated form.
+                 */
+                 if (strpos($key, 'AccessFu.properties') !== false) {
+                     return true;
+                 }
```
pascalchevrel: Again, I think this would be an improvement with a common comment above
```php
if (Strings::inString($key, ['region.properties', 'defines.inc', 'AccessFu.properties'])) {
    return true;
}
```

app/models/consistency.php
```diff
@@ -0,0 +1,79 @@
...
+);
+
+$target_locales_list = Utils::getHtmlSelectOptions(
+    Project::getRepositoryLocales($repo),
+    $locale
+);
+
+/*
+    Since this is a very heavy view for Aurora (default repo), we display
+    an introductory text when reaching the page and use a paramter to
```
pascalchevrel: parameter

app/models/consistency.php
```diff
@@ -0,0 +1,79 @@
...
+    determine if we actually want to display the strings.
+*/
+$display_results = isset($_GET['display']) ? true : false;
```
pascalchevrel: isset() returns a boolean, you can just write:
```php
$display_results = isset($_GET['display']);
```

app/models/consistency.php
```diff
@@ -0,0 +1,79 @@
...
+$reference_locale = Project::getReferenceLocale($repo);
+
+if ($display_results) {
```
pascalchevrel: actually, since you use $display_results only once, you could just as well put
```php
if (isset($_GET['display'])) {
```

app/models/consistency.php
```diff
@@ -0,0 +1,79 @@
...
+    // Get non-empty strings for reference and locale in this repository
+    $source_strings = array_filter(Utils::getRepoStrings($reference_locale, $repo), 'strlen');
```
pascalchevrel: You probably don't need to use a callback function like strlen because by default array_filter() filters results evaluating to false and an empty string evaluates to false.

app/classes/Transvision/Consistency.php
```diff
@@ -0,0 +1,121 @@
...
+             if (strlen($value) <= 1) {
+                 return true;
+             }
+
+             // Exclude CSS rules, "widht:" or "height:"
+             if (strpos($value, 'width:') !== false || strpos($value, 'height:') !== false) {
+                 return true;
+             }
+
+             // Exclude CSS widht values like '38em'
```
pascalchevrel: width

app/models/consistency.php
```diff
@@ -0,0 +1,79 @@
...
+    $target_strings = array_filter(Utils::getRepoStrings($locale, $repo), 'strlen');
+
+    if (! empty($source_strings)) {
+        $duplicated_strings_english = Consistency::findDuplicates($source_strings);
+        $duplicated_strings_translation = Consistency::findDuplicates($target_strings);
```
pascalchevrel: You are using this temporary array only once line 37. If you do line 37 like that:
```php
    // Get the strings duplicated in English but not in the localization
    $missing_duplicates = array_diff_key(
        $duplicated_strings_english, 
        Consistency::findDuplicates($target_strings)
    );
```
You save 2MB of memory

app/models/consistency.php
```diff
@@ -0,0 +1,79 @@
...
+                $available_translations = array_unique($available_translations);
+
+                $inconsistencies[] = [
+                    'source' => $current_source,
+                    'target' => $available_translations,
+                ];
+            }
+        }
+    } else {
+        $strings_number = 0;
```
pascalchevrel: The else clause adds complexity, especially since there are many ifs before and I had to look closely to see which if it was linked to. I think that since this is the default value, it should be set up early like line 23 with a comment indicating how it will be used in the view.

app/models/consistency.php
```diff
@@ -0,0 +1,79 @@
...
+*/
+$display_results = isset($_GET['display']) ? true : false;
+$reference_locale = Project::getReferenceLocale($repo);
+
+if ($display_results) {
+    // Get non-empty strings for reference and locale in this repository
+    $source_strings = array_filter(Utils::getRepoStrings($reference_locale, $repo), 'strlen');
+    $target_strings = array_filter(Utils::getRepoStrings($locale, $repo), 'strlen');
+
+    if (! empty($source_strings)) {
```
pascalchevrel: I would test for true and return early, this way you gain an indentation level on all the code that follows and that makes it easier to read: 
```php
// We are going to show the placeholder text and not do any calculation
if (! $display_results) {
    return;
}

// Get non-empty strings for reference and locale in this repository
$source_strings = array_filter(Utils::getRepoStrings($reference_locale, $repo));
$target_strings = array_filter(Utils::getRepoStrings($locale, $repo));
…
```

app/models/consistency.php
```diff
@@ -0,0 +1,79 @@
...
+        $duplicated_strings_translation = Consistency::findDuplicates($target_strings);
+
+        // Filter English duplicates before comparing them with the localization
+        $duplicated_strings_english = Consistency::filterStrings($duplicated_strings_english, $repo);
+
+        // Get the strings duplicated in English but not in the localization
+        $missing_duplicates = array_diff_key($duplicated_strings_english, $duplicated_strings_translation);
+    }
+
+    if (! empty($duplicated_strings_english)) {
```
pascalchevrel: I would again return early and gain an indentation level on the following code:
```php

// No duplicated strings to analyze, get back to the view.
if (empty($duplicated_strings_english)) {
    return;
}
```

flodolo:
Thanks, I think I've addressed all review comments. I went with a slightly different approach for the inString method, adding the possibility to match any or all needles.

app/classes/Transvision/Consistency.php
```diff
@@ -0,0 +1,115 @@
...
+    {
+        $duplicates = [];
+        asort($strings_array);
+
+        $previous_key = '';
+        $previous_value = '';
+        foreach ($strings_array as $key => $value) {
+            if (strcasecmp($previous_value, $value) === 0) {
+                $duplicates[$previous_key] = $previous_value;
+                $duplicates[$key] = $strings_array[$key];
```
flodolo: This should simply be
```PHP
$duplicates[$key] = $value;
```

app/models/consistency.php
```diff
@@ -0,0 +1,85 @@
...
+    $duplicated_strings_english = Consistency::findDuplicates($source_strings);
+
+    // Filter English duplicates before comparing them with the localization
+    $duplicated_strings_english = Consistency::filterStrings($duplicated_strings_english, $repo);
+
+    // Get the strings duplicated in English but not in the localization
+    $missing_duplicates = array_diff_key(
+        $duplicated_strings_english,
+        Consistency::findDuplicates($target_strings)
+    );
```
flodolo: I was trying to understand why a string supposed to be duplicated would only display one result in the translation column, and realized that the logic is flawed and I need to find a completely different approach.

Example: there are two strings for 'closed' in English, but one is translated as 'chiusa', the other one as 'chiuso', so they're not duplicated in Italian.

app/models/consistency.php
```diff
@@ -0,0 +1,85 @@
...
+
+// Get non-empty strings for reference and locale in this repository
+$source_strings = array_filter(Utils::getRepoStrings($reference_locale, $repo));
+$target_strings = array_filter(Utils::getRepoStrings($locale, $repo));
+
+if (! empty($source_strings)) {
+    $duplicated_strings_english = Consistency::findDuplicates($source_strings);
+
+    // Filter English duplicates before comparing them with the localization
+    $duplicated_strings_english = Consistency::filterStrings($duplicated_strings_english, $repo);
```
flodolo: Should filter strings before searching for duplicates. For example "closed" is available in two strings, but one needs to be filtered out, so it's not actually duplicated.

app/classes/Transvision/Consistency.php
```diff
@@ -0,0 +1,115 @@
...
+
+         // Determine if a string should be excluded
+         $ignore_string = function ($key, $value) use ($repository_type) {
+             // Ignore strings of length 1 (e.g. accesskeys) and empty strings.
+             if (strlen($value) <= 1) {
+                 return true;
+             }
+
+             // Exclude CSS rules, "width:" or "height:"
+             if (Strings::inString($key, ['width:', 'height:'])) {
```
flodolo: `$value`, not `$key`

app/models/consistency.php
```diff
@@ -0,0 +1,79 @@
...
+    an introductory text when reaching the page and use a paramter to
+    determine if we actually want to display the strings.
+*/
+$display_results = isset($_GET['display']) ? true : false;
+$reference_locale = Project::getReferenceLocale($repo);
+
+if ($display_results) {
+    // Get non-empty strings for reference and locale in this repository
+    $source_strings = array_filter(Utils::getRepoStrings($reference_locale, $repo), 'strlen');
+    $target_strings = array_filter(Utils::getRepoStrings($locale, $repo), 'strlen');
```
pascalchevrel: Thinking about it, array_filter should be done in findDuplicates() and not in the model, that way you get it defined only once and you don't need to think that the data provided has to be filtered for empty strings if the method is used in another view

flodolo:
Completely changed the model's logic, I think it's sound now.

I filter out know problematic strings and then search for duplicated in English (was doing the opposite before). Then I traverse the array checking translations for each of these strings, and keep the ones with more than one translation.

flodolo:
Looking into the failing test

pascalchevrel:
The new version is much faster, I think that the splash screen is no longer useful if the view has no performance hit.

flodolo:
True, getting rid of it.

app/classes/Transvision/Strings.php
```diff
@@ -64,15 +64,36 @@ public static function endsWith($haystack, $needles)
...
+            } else {
+                $missing_needles++;
+            }
+        }
+
+        if ($match_all && $missing_needles == 0) {
+            return true;
+        }
+
+        return false;
```
pascalchevrel: That seems a bit complicated, also $total_needles isn't used. I would rather write that:
```php
    public static function inString($haystack, $needles, $match_all = false)
    {
        $match = 0;
        foreach ((array) $needles as $needle) {
            if (mb_strpos($haystack, $needle, $offset = 0, 'UTF-8') !== false) {
                // If I need to match any needle, I can stop at the first found
                if (! $match_all) {
                    return true;
                }
                $match++;
            }
        }

        return $match == count($needles);
    }
```

flodolo: Agreed, definitely more readable.

PR #649, opened 2016-02-25
Add a view for commonly unlocalized words

TheoChevalier:
I'd guess we want to move the JSON view under /api, but not sure where


app/inc/dispatcher.php
```diff
@@ -81,6 +81,15 @@
...
         $page_descr = 'Display a list of strings identical to English';
         break;
+    case 'unlocalized':
+        $controller = 'unlocalized_words';
+        $page_title = 'Commonly unlocalized words';
+        $page_descr = 'Display the list of the most common untranslated words across all languages';
+        break;
+    case 'unlocalized-json':
+        $controller = 'unlocalized_words';
+        $template = false;
```
flodolo: = alignment

app/models/unlocalized_words.php
```diff
@@ -0,0 +1,104 @@
+<?php
+namespace Transvision;
+
+// Build arrays for the search form, ignore mozilla_org
+$channel_selector = Utils::getHtmlSelectOptions(
+    $repos_nice_names,
```
flodolo: Either the code is wrong, or the comment above is (mozilla.org is included).

app/models/unlocalized_words.php
```diff
@@ -0,0 +1,104 @@
...
+
+$ref_locale = Project::getReferenceLocale($repo);
+
+/*
+    pspell helps getting rid of false positive results by keeping only valid
+    English words. The downside is that it’s filtering out 'jargon' words that
+    can be used in devtools or Mozilla-specific words.
+
+    This requires the following packages to be installed:
+    sudo apt-get install libpspell-dev php5-pspell aspell-en
```
flodolo: Does it mean we break badly for Windows?

app/models/unlocalized_words.php
```diff
@@ -0,0 +1,104 @@
...
+foreach ($strings_reference as $string_ref_id => $ref_words) {
+    $english_words = [];
+    $locale_words = $strings_locale[$string_ref_id];
+
+    /*
+        If the string is missing in the locale or has been copy pasted from
+        source (e.g. not translated), skip it
+    */
+    if (! isset($locale_words)
+        || ($ref_words == $locale_words && $locale != $ref_locale)) {
```
flodolo: I guess you should skip the entire view if `$locale == $ref_locale`?

app/models/unlocalized_words.php
```diff
@@ -0,0 +1,104 @@
...
+        continue;
+    }
+
+    /*
+        Remove punctuation characters from the strings then explode them into
+        words
+    */
+    $ref_words    = explode(' ',
+        preg_replace('/(?![.=-])\p{P}/u', '', $ref_words)
```
flodolo: I might be oversimplifying or missing something, but why not using `str_replace`?

app/views/unlocalized_words.php
```diff
@@ -0,0 +1,28 @@
...
+
+$content = "<table class='collapsable results_table {$search_id}'>\n" .
+            "  <tr class='column_headers'>\n" .
+            "    <th>English</th>\n" .
+            "    <th>Occurrences</th>\n" .
+            "    <th></th>\n" .
+            "  </tr>\n";
+
+foreach ($unlocalized_words as $string_id => $string_value) {
+    $link = "/?recherche={$string_id}&repo={$repo}&sourcelocale=en-US" .
```
flodolo: `$ref_locale` instead of `en-US`?

app/views/unlocalized_words.php
```diff
@@ -0,0 +1,28 @@
...
+            "&locale={$locale}&search_type=strings&whole_word=whole_word";
+
+    $content .= "  <tr class='{$search_id}'>\n" .
+                "    <td>{$string_id}</td>\n" .
+                "    <td>{$string_value}</td>\n" .
+                "    <td><a href='{$link}'>🔍</a></td>\n" .
```
flodolo: It probably makes sense to have some text here for accessibility, and eventually hide it with CSS.

app/models/unlocalized_words.php
flodolo: Never mind, misread the regex. Still, why do you need the part before `\p{P}`?

app/models/unlocalized_words.php
TheoChevalier: Hm, does Transvision supports Windows?

TheoChevalier: I've seen some composer version of pspell, I'll to check if that works and the perfs are the same

app/models/unlocalized_words.php
TheoChevalier: I think Chris finds value in seeing the most common English words, that's why I added an exception

app/models/unlocalized_words.php
TheoChevalier: My level of understanding regexps is pretty low, so I think = and - can be removed, but not sure if . is used as a special character, or if it's meant to match with periods. If it's matching periods, we'd need it.

app/models/unlocalized_words.php
TheoChevalier: Looks like what I've found is just a wrapper for the package https://github.com/prodeveloper/pspell/blob/master/src/Chencha/Pspell/Validators/CheckPspellIsInstalled.php#L13

However, pspell should be available for Windows http://code.stephenmorley.org/php/enabling-pspell-in-xampp-on-windows/

app/models/unlocalized_words.php
TheoChevalier: looks like '/\p{P}/u' is returning quite more results, so that looks better

app/models/unlocalized_words.php
```diff
@@ -0,0 +1,104 @@
...
+
+$ref_locale = Project::getReferenceLocale($repo);
+$unlocalized_words = [];
+
+/*
+    pspell helps getting rid of false positive results by keeping only valid
+    English words. The downside is that it’s filtering out 'jargon' words that
+    can be used in devtools or Mozilla-specific words.
+*/
+if (! $pspell_link = \pspell_new('en_US', '', '', '', PSPELL_FAST)) {
```
TheoChevalier: @flodolo Is it the right way to do?

flodolo: I doubt it, it should throw an exception since the class is not available.

This should work:
```PHP
if (! class_exists('\pspell_new')) {
    // Do something
} else {
...
}
```

And I would display something to the user, I think we're giving a white page right now.

TheoChevalier:
I wonder why the functional test is failing on Travis while it's working locally…

app/models/unlocalized_words.php
TheoChevalier: I would redirect to a 404, but we don't have one yet :-/ In the meantime, we could just add text in the die()?

flodolo: Still a white page, pretty awful. Anyhow, before doing massive changes, let's see what Pascal thinks.

TheoChevalier:
Let's fix the test separately as it seems tricky.

PR #651, opened 2016-02-29
Fixes in health status view + removing repocomparison.php

TheoChevalier:


app/models/health_status.php
```diff
@@ -260,7 +261,7 @@ function ($entity) use ($path) {
     $stats = Health::getStats($projects);
     $translated = $stats['translated'];
     $reference = $stats['total'];
-    $completion = round(($translated / $reference) * 100, 2);
+    $completion = $reference === 0 ? 0 : round(($translated / $reference) * 100, 2);
```
pascalchevrel: ternary assignation is cryptic to me there :) Mind adding a comment explaining what you are doing? Thanks

flodolo: Also not sure about the identical operator, equivalence should be enough?

Issue #655, opened 2016-03-02
Some direct links to strings are leading to blank pages


TheoChevalier:
Just happened to me, didn't dig into it yet.

From this search, https://transvision.mozfr.org/?recherche=ail%2Fchrome%2Fmessenger%2Fmime.properties&repo=beta&sourcelocale=en-US&locale=fr&search_type=strings_entities
Click on entity link, for instance https://transvision.mozfr.org/?sourcelocale=en-US&locale=fr&repo=beta&search_type=entities&recherche=mail/chrome/messenger/mime.properties:1008

-> blank page

flodolo:
Interesting, no error on log or on video locally either :open_mouth:

pascalchevrel:
It seem to only happen when the entity ends with a number: 
mail/chrome/messenger/mime.properties:1008

flodolo:
I could be wrong but I think we fail around [here](https://github.com/mozfr/transvision/blob/master/app/inc/router.php#L3-L10) (very early). But in parse_url?

pascalchevrel:
yes, parse_url returns false when the url is not parsable, which happens when we end it with :1008, maybe it thinks that it is a port. I have a local patch that seems to work, at least with the dev server, will investigate more tomorrow

PR #656, opened 2016-03-02
Issue #655: escape semi-colon in query string to make the url parsable

pascalchevrel:


flodolo:
Looks good, also good idea to log similar errors :+1:

Issue #658, opened 2016-03-03
[consistency] locale version showing missing spaces 


goofy-bz:
A number of words are concateneted wihout the usual space in between in the locale column.

![capture du 2016-03-03 20 36 19](https://cloud.githubusercontent.com/assets/858630/13506860/4a479288-e180-11e5-8b02-58ccb58e9769.png)

flodolo:
I can confirm. The code is [using](https://github.com/mozfr/transvision/blob/master/app/views/consistency.php#L48) `Utils::secureText` to output the text, I wonder if that's the issue (but it's weird such a different behavior between two strings so similar).

pascalchevrel:
The difference is that the first string has no line breaks while the second string has, my guess is that secureText removes the line breaks and doesn't replace them by spaces

flodolo:
Exactly, just realized that one version of the string doesn't have whitespaces but new lines
https://hg.mozilla.org/releases/l10n/mozilla-aurora/fr/file/default/dom/chrome/netError.dtd#l190

Not sure if we should change secureText or use a different method. I think secureText is used in several views.

pascalchevrel:
I think it makes sense to add it to secureText because when we have a multiline string, we don't want the security filtering function to output a string that is different from the original one. Currently we remove ascii characters below 32 but e could just replace /n with a space before that operation.

flodolo:
Agreed. We're probably broken in other views and never noticed it.

pascalchevrel:
yes, I just checked the onestring view which uses secureText and it's broken for this string. Patch coming.

pascalchevrel:
fixed, thanks Goofy!

PR #660, opened 2016-03-03
Issue  #658: fix missing spaces in views

pascalchevrel:


flodolo:
Tested and it looks good (pending php-cs fix).

Issue #669, opened 2016-03-04
Glossaire.sh error on beta site


pascalchevrel:
running glossaire on the beta site yields errors:
Silme couldn't extract data for /var/cache/transvision/hg//RELEASE_EN-US/COMMUN/browser
/var/cache/transvision/hg//RELEASE_EN-US/COMMUN/browser/extensions/pocket/locales/en-US/en-US: [Errno 2] No such file or directory: '/var/cache/transvision/hg//RELEASE_EN-US/COMMUN/browser/extensions/pocket/locales/en-US/en-US'
English strings have been updated.
./app/scripts/glossaire.sh: ligne 229 : [: false : nombre entier attendu comme expression
./app/scripts/glossaire.sh: ligne 229 : [: false : nombre entier attendu comme expression

I don't know why, I'll investigate on my local instance.

pascalchevrel:
ok, I can reproduce locally when I set checkrepo=false in the script

pascalchevrel:
```bash
                if [ "$checkrepo" = true ]
                then
                    updateLocale ${!repo_l10n} $locale $repo_name/$locale
                    updated_locale=$?
                fi
              ...
                if [ "$forceTMX" = true -o "$updated_english" = true -o "$updated_locale" -eq 1 ]
                then
                    buildCache $locale
```

updated_locale is only defined in the if clause if checkRepo=true, that's why the script fails when checkrepo=false.

flodolo:
It's actually initialized on [line 212](https://github.com/mozfr/transvision/blob/master/app/scripts/glossaire.sh#L212), right before this piece of code.

PR #671, opened 2016-03-05
glossaire.sh: initialize update_locales to 0, not false (fix issue #669)

flodolo:
Fix issue #669


pascalchevrel:
It also needs to be fixed lines 323 and 352.

flodolo:
Argh, thanks (updated).

PR #675, opened 2016-03-11
Improve table semantics using thead and tbody instead of a standard row as header

### LINKED ISSUE #670: [unlocalized strings] Sanitize English words before displaying, fix first row style
Realized on my demo server where the packages weren't installed yet, HTML code was displayed as part of the English string (and iframe was displayed).

We shouldn't also displaying the first row as an header (I had the same issue with the consistency table, so there's a rule do display the right color).

flodolo:
Strip tags in unlocalized view, that also improves results (e.g. for words like "tag", "file", etc.).
Also fixes #670

### DIFF
--- a/app/views/gaia.php
+++ b/app/views/gaia.php
@@ -100,15 +100,15 @@ $sections = [
 // Overview of string count for the locale
 $overview = function ($section_title, $columns, $rows, $anchor) {
     // Titles
-    $html = '<table id="' . $anchor . '">'
-       . '<tr>'
-       . '<th colspan="3">' . $section_title . '</th>'
-       . '</tr>'
-       . '<tr>';
+    $html = "
+        <p class='section_title'>{$section_title}</p>
+        <table id='{$anchor}'>
+          <thead>
+            <tr class='column_headers'>";
     foreach ($columns as $key => $value) {
         $html .= '<th>' . $value . '</th>';
     }
-    $html .= '</tr>';
+    $html .= "</tr>\n</thead>\n<tbody>\n";
 
     // Rows
     foreach ($rows as $key => $row) {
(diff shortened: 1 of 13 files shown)


app/views/gaia.php
```diff
@@ -217,16 +218,21 @@
...
-       . '</tr>'
-       . '<tr>'
-       . '<th>Key</th>'
-       . '<th>' . $repo_one . '</th>'
-       . '<th>' . $repo_two . '</th>'
-       . '</tr>';
 
+$section_title = "Strings that have changed significantly in English between {$repo_one} and {$repo_two} but for which the entity name didn’t change";
+$table = "
+    <p class='section_title'>{$section_title}</p>
```
pascalchevrel: If it's a title maybe it should use a <hx> tag ?

flodolo: Not completely sure: it's not an actual title, more a subtitle, and before it was even more confusing (part of the table).

schermata 2016-03-11 alle 13 54 01

flodolo: (it might make sense to change the class name, e.g. section_description or similar)

pascalchevrel: or many a <th> that would span all columns if it's part of the table, or a <caption> tag?

flodolo: Let me check how `<caption>` is displayed.

flodolo: Tried but it would require quite a few changes, especially for smaller viewports (it breaks quite badly).

flodolo:
Updated variable/CSS class name in a separate commit. Not sure if we should drop the bold too.

pascalchevrel:
oops, missed that it wasn't squashed yet :(

PR #681, opened 2016-03-11
Use two columns for QA views

### LINKED ISSUE #663: Menu : have 2 columns for the QA view
We have more and more QA views which is good for a QA tool. We have horizontal space available in the menu so let's split the QA column in two columns to accommodate the new views.

flodolo:
Fix #663, it should be good for a test run.

Tested with Chrome and Firefox, it seems to work without particular issue. I decided to go vertical on the smaller viewport though, I think the text would be too small with 2 columns.

### DIFF
--- a/app/views/templates/base.php
+++ b/app/views/templates/base.php
@@ -31,18 +31,20 @@ $links = '
     <li><a ' . ($url['path'] == 'downloads' ? 'class="selected_view" ' : '') . 'href="/downloads/" title="Download TMX files">TMX Download</a></li>
   </ul>
 </div>
-<div class="linkscolumn">
+<div class="linkscolumn" id="qa_column">
   <h3>QA Views</h3>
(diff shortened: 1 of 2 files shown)


pascalchevrel:
The failure is a Travis bug

pascalchevrel:
It works well, I light have gone the road of using real css columns instead of floats (https://developer.mozilla.org/fr/docs/Web/CSS/columns) but this version is good for me :+1:

flodolo:
I don't think you can have columns with different widths? You could create 4 columns, but then you couldn't center the "QA views" title, and if you move the titles "out" of the columns you'll have issues with smaller viewports (title are not preceding their links).

flodolo:
Waiting for Travis to re-run, then I'll merge. If we remove the "logo", we can also recover some space at the top of the container.

Issue #223, opened 2014-04-08
[enhancement] have a specific friendly-looking logo/icon/favicon


goofy-bz:
this is a minor request. There seems to be no logo any longer since https://github.com/mozfr/transvision/commit/725d9787d004491fd100273dedae25f50d2b9373#diff-581b5e9ce03a747504f4a3e8d5e14ffd

Would you please consider having a new one?
Maybe it should be done by a call to graphic arts contribution on Mozilla l10n groups  (so it may arise extra interest for Transvision)

flodolo:
What about using the new [Community Design](https://github.com/mozilla/Community-Design/) group? They seem to be doing a pretty good job

flodolo:
Side thought: should it be only a favicon or also a full logo to use somewhere in the page? I'd like the idea of having both.

pascalchevrel:
Ideally a logo that could be declined as a favicon, I would also like something that can be printed on a Tshirt.

pascalchevrel:
I didn't know about the Community Design group, apparently they set up their GitHub repo while I was on medical leave :) I just filed an issue there https://github.com/mozilla/Community-Design/issues/34

pascalchevrel:
Elio Qoshi created a great set of logos and all the graphics are there:
https://github.com/mozilla/Community-Design/tree/master/Transvision

Taking this bug for the implementation

PR #682, opened 2016-03-11
Issue #223: Add our logo

pascalchevrel:
- add the logo in svg format and make it scale via CSS for screen widths
- add a svg favicon for Firefox and a 16x16 simple png one for other browsers
- store all the logos in the web/img/logo/ folder
- Credit Elio Qoshi and the Mozilla Community Design team

resources


pascalchevrel:
@flodolo and @TheoChevalier what do you think?

flodolo:
I like it. Some notes:
* Should we get rid of ´$title´ in `dispatcher.php`? I didn't check in full details, but I don't think it's used anywhere.
* Should we put a link to the home page on the logo? I definitely used the old title to go back.

flodolo:
Since we get rid of the "Transvision" fixed badge at the top, we can recover some top margin in the menu (if we want).

pascalchevrel:
I added the link back and removed $title

app/views/credits.php
```diff
@@ -26,6 +26,8 @@
 
 <p>Transvision is available on <a href="https://github.com/mozfr/transvision/">GitHub</a></p>
 
+<p>The Transvision logo was created by Elio Qoshi from the <a href="https://github.com/mozilla/Community-Design">Mozilla Community Design team</a></p>
```
flodolo: Looks like this sentence would need a period at the end (also the one before).

PR #685, opened 2016-03-13
Menu tooltip text default value is wrong

TheoChevalier:
Clearly not a critical patch…


flodolo:
Thanks, it remained around from #580

PR #686, opened 2016-03-14
Version 4.2 changelog

pascalchevrel:
* includes CSS fixes to the changelog page
* adds a new regression tag


app/models/changelog_data.php
```diff
@@ -52,9 +52,173 @@
...
+        ],
+        [
+            'message' => ['Fixes in health status view'],
+            'section' => ['code'],
+            'type'    => ['better'],
+            'commit'  => ['a3c97f25f97b74e16d5fa34c99baa2b864130612'],
+            'authors' => ['tchevalier'],
+        ],
+        [
+            'message' => ['Use prefer-dist unstead of prefer-source for Composer in Travis'],
```
TheoChevalier: typo: instead

app/models/changelog_data.php
```diff
@@ -52,9 +52,173 @@
...
+        ],
+        [
+            'message' => ['Improve security of our GitHub Web hook'],
+            'section' => ['code'],
+            'type'    => ['better'],
+            'issues'  => [680],
+            'authors' => ['pascal'],
+        ],
+        [
+            'message' => ['Use more meaningful name for section description'],
```
flodolo: We should drop this one. It's part of #675 , it just wasn't squashed before merge.

app/models/changelog_data.php
```diff
@@ -52,9 +52,173 @@
...
+            'section' => ['code'],
+            'type'    => ['better'],
+            'commit'  => ['dd2b843f0ee4a730e013b3a64c3c858044e04ae2'],
+            'authors' => ['flod'],
+        ],
+        [
+            'message' => ['[unlocalized strings] Sanitize English words before displaying, fix first row style'],
```
flodolo: This is the same as 'Use table head instead of a special row for column headers', I'd keep that one.

web/style/changelog.css
```diff
@@ -75,35 +76,35 @@
     left: 0;
     line-height: 1.1;
     padding: 4px 2px 4px 13px;
-    text-shadow: 1px 1px 0 #fff;
     text-transform: uppercase;
     width: 92px;
     margin-right:4px;
+    border-color: #555555;
```
flodolo: User short version: `#555`

web/style/changelog.css
```diff
@@ -75,35 +76,35 @@
...
+    background-color: #96e196;
 }
 
 .change {
     background-color: #e9cfff;
-    border-color: #8106e0;
+}
+
+.regression {
+    color: #ffffff;
```
flodolo: User short version: `#fff`

Issue #692, opened 2016-03-15
In top menu some pages don't get highlighted when you are on the specific page


pascalchevrel:
Happens for Home,  Unchanged Strings and Unlocalized Words.

I have a patch that also simplifies the code.

PR #693, opened 2016-03-15
Issue #692 - Fix highlighting of menu items

pascalchevrel:
Replace hardcoded HTML <li> tag with their ternary check to add a CSS class by a closure that generates the link in an heredoc variable.
This way we are sure we don't have typos and if we need to change the HTML output, it's not duplicated.

### DIFF
--- a/app/views/templates/base.php
+++ b/app/views/templates/base.php
@@ -21,41 +21,68 @@ if (isset($css_include)) {
     $css_include = $base_css;
 }
 
-$links = '
+/*
+    This is a closure to build a list item containing the
+    link to the page. The page we are on gets a CSS class
+    of selected_view added.
+*/
+$li_link = function ($page, $title, $text) use ($url, $urls) {
+    $link = array_search($page, $urls);
+
+    if ($link != '/') {
+        $link = "/{$link}/";
+    }
+
+    return '<li><a ' . ($url['path'] == $link ? 'class="selected_view" ' : '')
+           . 'href="' . $link . '" '
+           . 'title="' . $title . '">'
+           . $text . '</a></li>';
+};
+
+/*
+    The t2t page is a legacy page without an entry in the $urls array.
+*/
+$li_t2t = '<li><a ' . (isset($_GET['t2t']) ? 'class="selected_view" ' : '')
+       . 'href="/?sourcelocale=' . $source_locale . '&locale=' . $locale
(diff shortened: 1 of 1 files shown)


app/views/templates/base.php
```diff
@@ -21,41 +21,64 @@
...
+/*
+    This is a closure to build a list item containing the
+    link to the page. The page we are on gets a CSS class
+    of selected_view added.
+*/
+$li_link = function ($page, $title, $text) use ($url, $urls) {
+    $link = array_search($page, $urls);
+
+    return '<li><a ' . ($url['path'] == $link ? 'class="selected_view" ' : '')
+           . 'href="/' . str_replace('/', '', $link) . '" '
```
flodolo: The only case where `$link` contains a slash is root, maybe it would clearer to just examine that case before the return statement and use `$link` directly
```PHP
$link = array_search($page, $urls);
if ($link != '/') {
    $link = "/{$link}";
}
```

app/views/templates/base.php
```diff
@@ -21,41 +21,64 @@
...
+    {$li_link('consistency', 'Translation Consistency', 'Translation Consistency')}
+    {$li_link('unchangedstrings', 'Display all strings identical to English', 'Unchanged Strings')}
+    {$li_link('unlocalized', 'Display common words remaining in English', 'Unlocalized Words')}
   </ul>
   <ul>
-    <li><a ' . ($url['path'] == 'channelcomparison' ? 'class="selected_view" ' : '') . 'href="/channelcomparison/" title="Compare strings from channel to channel">Channel Comparison</a></li>
-    <li><a ' . ($url['path'] == 'gaia' ? 'class="selected_view" ' : '') . 'href="/gaia/" title="Compare strings across Gaia channels">Gaia Comparison</a></li>
-    <li><a ' . ($url['path'] == 'showrepos' ? 'class="selected_view" ' : '') . 'href="/showrepos/" title="Check the health status of locales">Health Status Overview</a></li>
-    <li><a ' . ($url['path'] == 'productization' ? 'class="selected_view" ' : '') . 'href="/productization/" title="Show productization aspects">Productization</a></li>
+    {$li_link('channelcomp', 'Compare strings from channel to channel', 'Channel Comparison')}
```
flodolo: between channels?

flodolo:
One possible suggestion, but it looks good and works as expected :+1:

Issue #697, opened 2016-03-16
Consistency: add filters to check only strings for group of products


flodolo:
I currently have a WIP patch, but it should be doable: basically add a filter to only check strings belonging to Firefox * (desktop and Android), Thunderbird, SeaMonkey.

PR #698, opened 2016-03-16
Consistency view: add filters for components

flodolo:
Fixes issue #697

### DIFF
--- a/app/models/consistency.php
+++ b/app/models/consistency.php
@@ -13,6 +13,29 @@ $target_locales_list = Utils::getHtmlSelectOptions(
     $locale
 );
 
+$available_filters = [
+    'all'         => 'All products',
+    'firefox'     => 'Firefox for desktop and Android',
+    'seamonkey'   => 'SeaMonkey',
+    'thunderbird' => 'Thunderbird',
+];
+
+if (isset($_GET['filter'])) {
+    $selected_filter = isset($available_filters[$_GET['filter']])
+        ? Utils::secureText($_GET['filter'])
+        : 'all';
+} else {
+    $selected_filter = 'all';
+}
+$filter_visibility = Project::isDesktopRepository($repo)
+    ? ''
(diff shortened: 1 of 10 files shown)


flodolo:
I've also switched the search link to use perfect match. I remember not using it because the perfect match was broken, but now it seems to work almost everytime and it's more useful than a generic search.

app/models/consistency.php
```diff
@@ -13,6 +13,29 @@
...
+$available_filters = [
+    'all'         => 'All products',
+    'firefox'     => 'Firefox for desktop and Android',
+    'seamonkey'   => 'SeaMonkey',
+    'thunderbird' => 'Thunderbird',
+];
+
+if (isset($_GET['filter'])) {
+    $selected_filter = isset($available_filters[$_GET['filter']])
+        ? $_GET['filter']
```
pascalchevrel: I think I'd prefer that you pass this $_GET parameter through ```Utils::secureText()``` because even though it's just used in a ternary test later in ```Utils::getHtmlSelectOptions()```, if we later change that function and use the string directly that could cause a XSS.

app/models/consistency.php
```diff
@@ -13,6 +13,29 @@
...
+        : 'all';
+} else {
+    $selected_filter = 'all';
+}
+$filter_visibility = in_array($repo, Project::getDesktopRepositories())
```
pascalchevrel: I grepped the repo and we use this ```if (in_array($repo, Project::getDesktopRepositories())) ''' pattern already 4 times, maybe we should have a Project::isDesktopRepository() method? (not necessarily in this PR, I can do that in a separate PR if you are busy)

pascalchevrel:
apart from a couple of nits, LGTM

flodolo:
Added changes in a separate commit if you want to double check

pascalchevrel:
ok for me :+1:

PR #699, opened 2016-03-16
Add an api to transliterate serbian cyrillic

SkySymbol:


app/models/api/transliterate.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+$t = Transliterator::create('Serbian-Latin/BGN');
+$translitareted_string = $t->transliterate($request->parameters[2]);
+
+return $json = ['translitareted_string' => $translitareted_string];
```
flodolo: Typo in the variable name. 

Also I don't think the extension is enabled by default (but didn't have a chance to test this yet). If that's the case, you should check if the class is available before trying to use it.

SkySymbol:
Thanks. I'm also wondering if we should add a more generic translieration api, like
``transliterate/serbian/xxx`` instead of ``translieration/xxx``

TheoChevalier:
Thanks for the patch Vincent, here's some general feedback and some use cases after a quick chat I had with Pascal:
- We could allow users to download transliterated TMX (currently, it would be sr-latn) on the TMX download view
- API should be neutral, so that we could support other languages in the future
- We could use this API on the main search view. For instance, we could display a "Transliterate" button for Serbian only that would replace the Cyrillic results with transliterated Latin (by putting both strings in each cell and using display: none to switch between them)

We might also ask Vanja (one of our Serbian localizer) for his feedback https://mozillians.org/u/civas/

app/models/api/transliterate.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+// Check that the class exists before trying to use it
+if (! class_exists('Transliterator')) {
+    return $json = ['error' => 'class Transliterator not available'];
```
TheoChevalier: I think you can set $request->error then use this function here https://github.com/mozfr/transvision/blob/master/app/classes/Transvision/API.php#L309

SkySymbol:
Yes, i focused the API in this PR because I think it's a good starting point. So we can use this API afterward to add features.

> API should be neutral, so that we could support other languages in the future

This why I thought about this api path ``transliterate/serbian/xxx``

app/classes/Transvision/API.php
```diff
@@ -299,6 +299,11 @@ private function isValidServiceCall($service)
                     No user-defined variables = nothing to check
                 */
                 break;
+            case 'transliterate':
+                if (! $this->verifyEnoughParameters(3)) {
```
TheoChevalier: api/v1/transliterate/serbian/ throws an error "Undefined offset: 3 in /home/theo/moz/transvision/app/models/api/transliterate.php on line 28" We should ask for 4 parameters here

app/models/api/transliterate.php
```diff
@@ -0,0 +1,30 @@
...
+switch ($request->parameters[2]) {
+    case 'serbian':
+        $transliterated_locale = 'Serbian-Latin/BGN';
+        break;
+
+    default:
+        break;
+}
+
+if (! $transliterated_locale) {
```
TheoChevalier: this is throwing a warning when calling the API with a wrong locale, we should use instead:
if (! isset($transliterated_locale))

app/models/api/transliterate.php
```diff
@@ -0,0 +1,30 @@
...
+    $request->error = 'Wrong language';
+    $json = $request->invalidAPICall();
+
+    return;
+}
+
+$t = Transliterator::create($transliterated_locale);
```
TheoChevalier: Let’s avoid single letter variables, $transliterator ?

tests/functional/api.php
```diff
@@ -21,6 +21,7 @@
     ['v1/suggestions/central/en-US/fr/bookmark/?max_results=2', 200, '["Bookmark","Bookmarks"]'],
     ['v1/tm/central/en-US/fr/Bookmark/?max_results=3&min_quality=80', 200, '[{"source":"Bookmark","target":"Marquer cette page","quality":100},{"source":"Bookmark","target":"Marque-page","quality":100},{"source":"Bookmarks","target":"Marque-pages","quality":88.89}]'],
     ['v1/tm/global/fr/en-US/Ouvrir/', 200, '[{"source":"Ouvrir dans le Finder","target":"Find in Finder","quality":28.57},{"source":"D\u00e9couvrez comment ouvrir une fen\u00eatre de navigation priv\u00e9e","target":"Learn how to open a private window","quality":8.77}]'],
+    ['v1/transliterate/serbian/%D0%A1%D1%80%D0%B1%D0%B8%D1%98a/', 200, '{"transliterated_string":"SrbiJa"}'],
```
TheoChevalier: We can also check this one:
['v1/transliterate/foo/bar/', 400, '{"error":"Wrong language"}'],

Also, current test is failing on my Ubuntu VM:
* Expected: {"transliterated_string":"SrbiJa"}
* Received: {"transliterated_string":"Srbija"}

Probably related to ICU version, not sure we can fix that :-/
I’m using php5-intl_5.6.11+dfsg-1ubuntu3.1_amd64.deb

TheoChevalier:
Ok, focusing on API first sounds good to me :)

A few details left, and I think we’ll be good.

I think we’ll also want to update https://github.com/mozfr/transvision/wiki/JSON-API

TheoChevalier:
Last thing, turns out php5-intl is not listed in dependencies, let’s add it https://github.com/mozfr/transvision/blob/master/README.md#dependencies

SkySymbol:
Thank you, all your points make sense. We can't check ['v1/transliterate/foo/bar/', 400, '{"error":"Wrong language"}'] because the method ``invalidAPICall()`` returns a 302 redirect and ``file_get_contents`` does not appear to  follow redirects (I don't why) so the test won't get the 400 error code.

> Also, current test is failing on my Ubuntu VM:
    Expected: {"transliterated_string":"SrbiJa"}
    Received: {"transliterated_string":"Srbija"}

I have the same bug on my debian box (php5-intl-5.6.19+dfsg-2)

TheoChevalier:
> Thank you, all your points make sense. We can't check ['v1/transliterate/foo/bar/', 400, '{"error":"Wrong language"}'] because the method invalidAPICall() returns a 302 redirect and file_get_contents does not appear to follow redirects (I don't why) so the test won't get the 400 error code.

What’s the error you get? When I add `['v1/transliterate/foo/bar/', 400, '{"error":"Wrong language"}'],` to api.php, it works for me.

> I have the same bug on my debian box (php5-intl-5.6.19+dfsg-2)
Ok, so maybe we should run this test on Travis only for now (guessing we could use similar trick https://github.com/mozfr/transvision/blob/master/tests/units/Transvision/Utils.php#L314 )

SkySymbol:
Well, I've just tried and it works now. I don't know why but when I first tested the test caught only the 302 redirect so I got:

```
HTTP return code error: http://localhost:8083/api/v1/transliterate/foo/bar/
* Expected: 400
* Received: 302
```

But it's ok now. I'll add the test and fix the conflicts asap ;-)

README.md
```diff
@@ -33,6 +33,8 @@ The Transvision team uses Git and GitHub for both development and issue tracking
 * Composer (Dependency Manager for PHP, https://getcomposer.org/)
 * mercurial, git, svn to check out data sources (only for a Full installation for production)
 * php5-xsl and GraphViz packages for generating the documentation with [phpDocumentor][]
+* php5-intl for the transliteration feature
+* libpspell-dev, php5-pspell and aspell-en packages for running spell checking in English on [Unlocalized words view][]
```
pascalchevrel: This line should be removed

app/classes/Transvision/API.php
```diff
@@ -299,6 +299,11 @@ private function isValidServiceCall($service)
                     No user-defined variables = nothing to check
                 */
                 break;
+            case 'transliterate':
```
pascalchevrel: this case should be before the 'version' case, the services in the switch are sorted alphabetically.
Also, there should be a comment indicating an API call example.

app/controllers/api.php
```diff
@@ -53,6 +53,9 @@
     case 'versions':
         $json = include MODELS . 'api/versions.php';
         break;
+    case 'transliterate':
```
pascalchevrel: again, case not sorted in the switch

app/controllers/api.php
```diff
@@ -53,6 +53,9 @@
...
+        include MODELS . 'api/transliterate.php';
```
pascalchevrel: You should do like other cases and create the $json variable there and not in the model:
```php
$json = include MODELS . 'api/transliterate.php';
```

app/models/api/transliterate.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+// Check that the class exists before trying to use it
```
pascalchevrel: mention here the dependency to php-intl

app/models/api/transliterate.php
```diff
@@ -0,0 +1,30 @@
...
+if (! class_exists('Transliterator')) {
+    $request->error = 'Class Transliterator not present';
```
pascalchevrel: Transliterator class not available.

app/models/api/transliterate.php
```diff
@@ -0,0 +1,30 @@
...
+    $json = $request->invalidAPICall();
```
pascalchevrel: That method returns a 400 HTTP code error, which is a user on the user's end. But here we have to deal with a server-side error, I suggest returning the 501 error code. That means updating this method to have a default value of 400 and deal with the 501 error (and update tests).
ex:
```php
$json = $request->invalidAPICall(501);
......
public function invalidAPICall($error = 400)
    {
        http_response_code($error);

        return ['error' => $this->error];
    }
```

app/models/api/transliterate.php
```diff
@@ -0,0 +1,30 @@
...
+
+    return;
+}
+
+switch ($request->parameters[2]) {
+    case 'serbian':
```
pascalchevrel: We shouldn't use locale names but locale codes. In this case, the locale code would be sr-Cyrl

app/models/api/transliterate.php
```diff
@@ -0,0 +1,30 @@
...
+
+    default:
+        break;
+}
+
+if (! isset($transliterated_locale)) {
+    $request->error = 'Wrong language';
+    $json = $request->invalidAPICall();
+
+    return;
```
pascalchevrel: since we don't support any fallback value to sr-Cyrl, that whole block should be returned directly in the switch case about in the default case.

app/models/api/transliterate.php
```diff
@@ -0,0 +1,30 @@
...
+}
+
+$transliterator = Transliterator::create($transliterated_locale);
+$transliterated_string = $transliterator->transliterate($request->parameters[3]);
+
+return $json = ['transliterated_string' => $transliterated_string];
```
pascalchevrel: I don't think there is value in having a key in the json returned, you can resturn transliterated string directly:
```php
return [$transliterator->transliterate($request->parameters[3])];
```

You should also sanitize the output to remove XSS injection on the API consumer side, example:
<сцрипт>алерт('XСС)<%2Fсцрипт>/

tests/functional/api.php
```diff
@@ -21,8 +21,16 @@
     ['v1/suggestions/central/en-US/fr/bookmark/?max_results=2', 200, '["Bookmark","Bookmarks"]'],
     ['v1/tm/central/en-US/fr/Bookmark/?max_results=3&min_quality=80', 200, '[{"source":"Bookmark","target":"Marquer cette page","quality":100},{"source":"Bookmark","target":"Marque-page","quality":100},{"source":"Bookmarks","target":"Marque-pages","quality":88.89}]'],
     ['v1/tm/global/fr/en-US/Ouvrir/', 200, '[{"source":"Ouvrir dans le Finder","target":"Find in Finder","quality":28.57},{"source":"D\u00e9couvrez comment ouvrir une fen\u00eatre de navigation priv\u00e9e","target":"Learn how to open a private window","quality":8.77}]'],
-    ['v1/versions/', 200, '{"v1":"stable"}'],
 ];
+    // determine if the test is run on Travis to avoid the php-intl bug
```
pascalchevrel: it would probably be better to find a string that is not affected by the php-intl bug and avoid a conditional here.

tests/functional/api.php
```diff
@@ -21,8 +21,16 @@
...
+    if (getenv('TRAVIS')) {
+        array_push($paths, ['v1/transliterate/serbian/%D0%A1%D1%80%D0%B1%D0%B8%D1%98a/', 200, '{"transliterated_string":"SrbiJa"}']);
```
pascalchevrel: is there a reason to escape the characters and not use Србијa directly?

SkySymbol: Yes, using Србијa produces this warning :

```
PHP Warning:  file_get_contents(http://localhost:8083/api/v1/transliterate/sr-Cyril/Србијa): failed to open stream: HTTP request failed!  in /home/vincent/transvision/vendor/pchevrel/verif/src/pchevrel/Verif.php on line 113
```
Thank you for the review Pascal

app/models/api/transliterate.php
```diff
@@ -0,0 +1,27 @@
...
+// Requires php-intl
+if (! class_exists('Transliterator')) {
+    $request->error = 'Class Transliterator not available';
+    $json = $request->invalidAPICall(501);
+
+    return;
+}
+
+switch ($request->parameters[2]) {
+    case 'sr-Cyril':
```
pascalchevrel: should be sr-Cyrl not sr-Cyril

app/classes/Transvision/API.php
```diff
@@ -293,6 +293,13 @@ private function isValidServiceCall($service)
                 }
 
                 break;
+            case 'transliterate':
+            // ex: /api/v1/transliterate/serbian/Завиритe
```
pascalchevrel: url in comment should be /api/v1/transliterate/sr-Cyrl/Завиритe

app/models/api/transliterate.php
```diff
@@ -0,0 +1,27 @@
+<?php
+namespace Transvision;
+
+// Check that the class exists before trying to use it
+// Requires php-intl
```
pascalchevrel: Keep the comment on a single line

app/models/api/transliterate.php
```diff
@@ -0,0 +1,27 @@
...
+    default:
+    $request->error = 'Wrong language';
+
+    return $request->invalidAPICall(501);
+}
+
+$transliterator = \Transliterator::create($transliterated_locale);
+$transliterated_string = $transliterator->transliterate($request->parameters[3]);
+
+return $json = [$transliterator->transliterate(Utils::secureText($request->parameters[3]))];
```
pascalchevrel: don't create a $json variable, just return the array.

app/models/api/transliterate.php
```diff
@@ -0,0 +1,27 @@
...
+        break;
+
+    default:
+    $request->error = 'Wrong language';
+
+    return $request->invalidAPICall(501);
+}
+
+$transliterator = \Transliterator::create($transliterated_locale);
+$transliterated_string = $transliterator->transliterate($request->parameters[3]);
```
pascalchevrel: The variable in this line is not used later

pascalchevrel:
I am going to merge it as is and fix the issues I mentioned in a separate commit, this way your patch will be in the next release. Thanks :)

Issue #701, opened 2016-03-17
Reset advanced search options when suggesting similar strings


flodolo:
I've been using Transvision heavily these days to improve consistency, and noticed that we sometime suggest a string when there are no results, but the link leads to another empty result.

Example with [perfect_match](https://transvision.mozfr.org/?recherche=%3Cbutton+id%3D%27getMeOutOfHereButton%27%3E%26securityOverride.getMeOutOfHereButton%3B%3C%2Fbutton%3E&repo=central&sourcelocale=en-US&locale=it&search_type=strings&perfect_match=perfect_match) and [whole_word](https://transvision.mozfr.org/?recherche=Expected+identifier+but+found+%27%251%24S%27.&repo=aurora&sourcelocale=en-US&locale=it&search_type=strings&case_sensitive=case_sensitive&whole_word=whole_word). Case sensitive is probably safe-ish, but I think it makes sense to reset all of them.

PR #702, opened 2016-03-17
Reset advanced search options in links to similar results (fixes issue #701)

flodolo:


flodolo:
Update with one call to unset outside of the foreach cycle, not sure if you prefer them on one line.

pascalchevrel:
one line is OK, there is a tolerance to 120 characters lines in PSR

Issue #569, opened 2016-01-04
Allow localizers extracting raw translations they’ve pushed during a cycle


TheoChevalier:
I’ve been doing that manually for some time now ( Editing https://github.com/mozfr/transvision/blob/master/app/scripts/tmxmaker.py#L54 locally to only return the text, e.g. one string per line, without its ID). Doing so allows me to check all the strings with a grammar checker (mine is http://www.antidote.info/antidote for both French and English). This also works when localizers are using a simple spell checker.

Currently it’s a pain to check 30,000 strings with lots of false positive where I could only check what changed between two branches. I also don’t do it for each release because it’s time consuming. Helping automate this would help catching typos like http://hg.mozilla.org/l10n-central/fr/rev/6cc9480baf4b One could even think of running a Jenkins server with a spell checker.

The data I need are basically the green and orange strings exposed by Elmo on https://l10n.mozilla.org/source/diff/?from=661304c55fbc&to=d9c01cf0f9e9&repo=l10n-central%2Ffr but without string IDs.
In Transvision world that would be New strings + strings displayed in Channel Comparison.
We could expose this in download view if you think several people would be interested in this, or just exposing an API if you don’t want to clutter the UI.

What do you guys think?

flodolo:
I can see a use case for a view displaying the strings changed between two branches, like we do for Gaia, extending https://transvision.mozfr.org/channelcomparison/ with 'Added strings' and 'Removed strings'

On the other hand, I'm not completely sure of how many people would use a file with only these strings, especially in an automated way, given the number of false positives.

TheoChevalier:
A view would be perfect. So maybe we could make the Gaia view generic enough to work for any repo and remove the Channel comparison? + add a way to query any of these set of strings via the API? (Just a thought)

flodolo:
Making the Gaia view more generic would definitely work for me, there are other interesting things in there (e.g. changed English strings).

An API call would work for me.

TheoChevalier:
Definitely agree for English strings, that would be more consistent as currently we have to check https://transvision.mozfr.org/channelcomparison/?locale=en-US&chan1=central&chan2=aurora to have the same data and that’s less obvious.

pascalchevrel:
Talked with Théo today and it seems that improving the channelcomparison view would suffice (basically it only lacks added strings). I would say it's doable during the hackathon.

TheoChevalier:
Follow up, tags still seem to be interpreted:
[...]

PR #705, opened 2016-03-22
Highlight columns in Channel Comparison for quick grammar checking

TheoChevalier:
One of the missing pieces to fix issue #569. Menu would be another PR. I was thinking about it, and turns out that highlighting text is making it easier than copying values directly

I still need to figure out how to make the click events generic (e.g. not relying on IDs) so that we have a single call at the bottom of select_column.js that would work for, let's say, all tables with 'selectable' class.

Also need to figure out the appropriate UI (icon, button…)

### DIFF
--- a/web/js/select_column.js
+++ b/web/js/select_column.js
@@ -1,0 +1,32 @@
+function selectColumn(index, tableID) {
+    var columnSelector = '#' + tableID + ' tbody > tr > td:nth-child(' + (index + 1) + ')';
+    var cells = $(columnSelector);
+
+    // Clear existing selections
+    if (window.getSelection) { // all browsers, except IE before version 9
+        window.getSelection().removeAllRanges();
+    }
+
+    if (document.createRange) {
+        cells.each(function(i, cell) {
+            var rangeObj = document.createRange();
+            rangeObj.selectNodeContents(cell);
+            window.getSelection().addRange(rangeObj);
+        });
+
+    } else { // Internet Explorer before version 9
+        cells.each(function(i, cell) {
+            var rangeObj = document.body.createTextRange();
+            rangeObj.moveToElementText(cell);
+            rangeObj.select();
+        });
+    }
+}
+
+$(document).ready(function() {
+    $('.select_header').click(function() {
+        var columnNumber = $(this).index();
+        var tableID = $(this).closest('table').attr('id');
+        selectColumn(columnNumber, tableID);
+    });
+});
--- a/web/style/transvision.css
+++ b/web/style/transvision.css
@@ -296,6 +296,23 @@ th {
     padding: 0.5em;
 }
 
+th.select_header {
+    position: relative;
+}
+
+th.select_header span {
+    position: absolute;
+    color: #fff;
+    background-color: #1c75bc;
+    right: 5px;
+    top: 8px;
+    font-weight: normal;
+    font-size: 11px;
+    padding: 2px 5px;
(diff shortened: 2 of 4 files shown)


web/js/select_column.js
```diff
@@ -0,0 +1,41 @@
...
+    } else {
+        cells.each(function(i, cell) {
+            var rangeObj = document.body.createTextRange();
+            rangeObj.moveToElementText(cell);
+            rangeObj.select();
+        });
+    }
+}
+
+$(document).ready(function() {
```
flodolo: You can drastically simplify this: assign a class `select_header` to the columns you want to be selectable (right now you let select also the key column, which is not useful), and use
```JS
$(document).ready(function() {
    $('.select_header').click(function() {
        var columnNumber = $(this).index();
        var tableID = $(this).closest('table').attr('id');
        SelectColumn(columnNumber, tableID);
    });
});
```

web/js/select_column.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    if (document.createRange) {
+        cells.each(function(i, cell) {
+            var rangeObj = document.createRange();
+            rangeObj.selectNodeContents(cell);
+            window.getSelection().addRange(rangeObj);
+        });
+
+    // Internet Explorer before version 9
```
flodolo: To which part of the code is this comment related?

web/js/select_column.js
```diff
@@ -0,0 +1,41 @@
+function SelectColumn(index, tableId) {
+    var columnSelector = '#' + tableId + ' tbody > tr > td:nth-child(' + (index + 1) + ')';
+    var cells = $(columnSelector);
+
+    // clear existing selections
```
flodolo: nit: Clear

TheoChevalier:
Added a basic UI

web/style/transvision.css
```diff
@@ -296,6 +296,15 @@ th {
     padding: 0.5em;
 }
 
+th.select_header::after {
+    content: "Select";
+    font-weight: normal;
```
flodolo: I'm not a huge fan: 
- The existing label is not centered anymore.
- It's not available when table collapses (maybe that's OK).
- The cursor should change on hover.
- The text should probably say "Select column" for more clarity.

What about adding a real element in the HTML (e.g. a `<span>`) and use absolute positioning. We could also re-use the copy icon, and use "Select column" as alt/title

web/js/select_column.js
```diff
@@ -0,0 +1,32 @@
+function SelectColumn(index, tableID) {
```
flodolo: Nit: selectColumn

web/style/transvision.css
pascalchevrel: For the cursor, I suggest:
```css
cursor: cell;
```

pascalchevrel:
the en-US column for the "New strings added in fr between Aurora and Beta" section has no select button, I think it should both for consistency with the first table and because I can imagine some people interested in copy/pasting this data easily

TheoChevalier:
Thanks, updated

flodolo:
It looks good, I wonder if we should we worry about the completely collapsed table?

TheoChevalier:
IMO that’s a feature, not a bug (e.g. not getting clumsy UI when you don’t need it).
People using grammar checker on tablets, that would surprise me :)

flodolo:
I guess that makes sense :smile:

Issue #706, opened 2016-03-23
Set search type in the Search object


pascalchevrel:
I have a patch that allows defining the search type  in the object and that cleans up and simplify search_options.php and search_form.php.

PR #707, opened 2016-03-23
Issue #706: Set search type in the Search object

pascalchevrel:
New methods in Search class:
* setSearchType($type)
* getSearchType()
* getSearchTypes()
* getFormSearchOptions()
* setLocales()
* getLocale()
* getFormSearchOptions()
* getFormCheckboxes()

Other changes:
* Only one array $cookies storing all cookies values
* Suppression of several global variables
* Refactoring of search_options.php and search_forms.php to use lambdas and remove code duplication
* ShowResults::resultsTable() now takes the $search object as a parameter

Generally speaking, it makes our core code simpler and hopefully easier to make evolve.

### DIFF
--- a/app/classes/Transvision/Search.php
+++ b/app/classes/Transvision/Search.php
@@ -5,18 +5,22 @@ namespace Transvision;
  * Search class
  *
  * Allows searching for data in our repositories using a fluent interface.
- * Currently, only the regex part (definition of the search) is implemented.
+ * Currently, only the regex part (definition of the search) and defining the
+ * search options are implemented.
  * e.g.:
  * $search = (new Search)
  *     ->setSearchTerms('Bookmark this page')
  *     ->setRegexWholeWords(true)
  *     ->setRegexCaseInsensitive(true)
- *     ->setRegexPerfectMatch(false);
+ *     ->setRegexPerfectMatch(false)
+ *     ->setRepository('release')
+ *     ->setSearchType('strings')
+ *     ->setLocales(['en-US', 'fr']);
  */
 class Search
 {
     /**
-     * The trimmed string searched, we keep it   as the canonical reference
+     * The trimmed string searched, we keep it as the canonical reference
      * @var string
      */
     protected $search_terms;
@@ -59,6 +63,38 @@ class Search
     protected $repository;
 
     /**
+     * The type of search we will perform
+     * @var string
+     */
+    protected $search_type;
+
+    /**
+     * The different types of search we can perform
+     * @var array
+     */
+    protected $search_types = ['strings', 'entities', 'strings_entities'];
+
+    /**
+     * The different options associated to searches
+     * @var array
+     */
+    protected $form_search_options = [
+        'case_sensitive', 'perfect_match', 'repo', 'search_type', 't2t', 'whole_word',
+    ];
+
+    /**
+     * The different checkboxes for the search Form
+     * @var array
+     */
+    protected $form_checkboxes;
+
+    /**
+     * The different locales we will use in views
+     * @var array
+     */
+    protected $locales;
+
+    /**
      * We set the default values for a search
      */
     public function __construct()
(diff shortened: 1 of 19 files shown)


app/inc/search_options.php
```diff
@@ -12,34 +12,24 @@
 // Checkboxes states
 $check = [];
 
+// Search forms
+$form_checkboxes = array_diff($search->getFormSearchOptions(), ['repo', 'search_type']);
```
flodolo: Since the other options are defined in the class, it might make more sense to move this too inside the class.

app/classes/Transvision/Search.php
```diff
@@ -59,6 +59,26 @@ class Search
...
+    protected $search_type;
+
+    /**
+     * The different types of search we can perform
+     * @var array
+     */
+    protected $search_types = ['strings', 'entities', 'strings_entities'];
+
+    /**
+     * The different types of search we can perform
```
flodolo: A bit misleading: maybe "The different options associated to searches"?

app/classes/Transvision/Search.php
```diff
@@ -59,6 +59,26 @@ class Search
...
+     * @var array
+     */
+    protected $form_search_options = [
+        'case_sensitive', 'whole_word', 'perfect_match', 't2t', 'repo', 'search_type',
```
flodolo: I would also order both these arrays alphabetically, just to make them easier to read.

flodolo:
I won't have time to review in depth until tomorrow, but a quick test searching on the home page without changing any option generates a ton of warnings.
```
Notice: Undefined index: repo in /app/classes/Transvision/ShowResults.php on line 279

Notice: Undefined index: search_type in /app/classes/Transvision/ShowResults.php on line 284

etc.
```

pascalchevrel:
I am looking into the warnings

pascalchevrel:
This is ready for review, it does more than setting the search type in the Search object, it also sets the locales, form options and checkboxes. I have updated the initial comment to reflect the changes.

app/classes/Transvision/Search.php
```diff
@@ -267,4 +310,92 @@ public function getRepository()
...
+     *
+     * @return array Form checkboxes
+     */
+    public function getFormCheckboxes()
+    {
+        return $this->form_checkboxes;
+    }
+
+    /**
+     * Set the Locales we will use for the search
```
flodolo: nit: locales

app/classes/Transvision/ShowResults.php
```diff
@@ -270,18 +268,18 @@ public static function resultsTable($search_id, $search_results, $recherche, $lo
                      </thead>
                      <tbody>\n";
 
-        if (!$search_options['whole_word'] && !$search_options['perfect_match']) {
-            $recherche = Utils::uniqueWords($recherche);
+        if (! $search_object->isWholeWords() && ! $search_object->isPerfectMatch()) {
+            $recherche = Utils::uniqueWords($search_object->getSearchTerms());
```
flodolo: Since we're touching this piece of code, can we switch to the English name for the variable? ;-)

pascalchevrel: sure :)

app/classes/Transvision/Search.php
```diff
@@ -267,4 +310,92 @@ public function getRepository()
...
+
+        return $this;
+    }
+
+    /**
+     * Get the locales used for searching
+     *
+     * @return array Locale codes
+     */
+    public function getLocales()
```
flodolo: In the code the function is always used as `$search->getLocales()[0]$search->getLocales()[NUMBER]`, which is a bit confusing. I think it would be clearer to have calls like `$search->getLocales()[0]$search->getLocales('source')` to get the specific locale: `source`, ´target´, `extra`, `all` in case we need the entire array (I don't think we do).

flodolo:
Everything works fine on my VM, and can't spot any issue with the code besides the note about the getLocales() method :+1:

pascalchevrel:
Review comments addressed in a separate commit, feel free to squash and merge if it look ok.

flodolo:
I'm getting some warnings now, but I'm still trying to figure out where they come from, and if they were there before (and if it's something due to my local data).
```
 Notice: Undefined offset: 2 in /home/flodolo/github/transvision/app/classes/Transvision/ShowResults.php on line 302
```

flodolo:
Found the issue: `$locale2` is set at
https://github.com/pascalchevrel/transvision/blob/clean_up_search_option_code/app/inc/l10n-init.php#L17

If I'm on an Italian browser, and search for French, I get the error.

pascalchevrel:
This new commit fixed the warning by explicitly passing the page name to resultsTable() instead of trying to detect in the method if we are are on the 3locales view according to the number of locales which can be tricky.

flodolo:
Works for me.

Issue #708, opened 2016-03-25
Remove Venkman


flodolo:
We stopped [building Venkman](https://bugzilla.mozilla.org/show_bug.cgi?id=1029513) and we're removing the folder from l10n repos. 

We should remove references to it before the next release.

PR #709, opened 2016-03-25
Remove Venkman support

flodolo:
Fix issue #708

Note that the repository has to be manually removed from /TRUNK_EN_US folder in order to clean up symlinks.


pascalchevrel:
shouldn't special cases for venkman in glossaire.sh and setup.sh be also removed?

flodolo:
> shouldn't special cases for venkman in glossaire.sh and setup.sh be also removed?
Uhm, any reference in particular? Because both files are part of this PR, but ChatZilla is still there.

pascalchevrel:
forget it, for some reason the page didn't load completely and I didn't see the other changes.

PR #714, opened 2016-04-01
Add a 404 page, fixes #568

### LINKED ISSUE #568: Create a 404 page
We never took time to create a 404 page, we probably should.

TheoChevalier:


TheoChevalier:
Added a smaller, bolder image and removed locales with unsupported characters.

app/views/404.php
```diff
@@ -0,0 +1,10 @@
+<?php
+namespace Transvision;
+
+?>
+<div id="pagecontent">
+    <p>
+        You can use the menu at the top to browse the available pages. If you think that’s an error, please <a href="https://github.com/mozfr/transvision/issues/new">open an issue on GitHub</a>.
```
flodolo: Maybe: "If you think this URL should work and it's a bug..."

app/inc/dispatcher.php
```diff
@@ -2,7 +2,8 @@
 namespace Transvision;
 
 $template     = true;
-$page         = $api_url ? 'api' : $urls[$url['path']];
+$page         = isset($urls[$url['path']]) ? $urls[$url['path']] : 'notfound';
+$page         = $api_url ? 'api' : $page;
```
flodolo: Maybe more readable like this?
```PHP
if ($api_url) {
    $page = 'api';
} else {
    $page = isset($urls[$url['path']]) ? $urls[$url['path']] : 'notfound';
}
```

flodolo:
I'm a bit torn on the image size: I would like it smaller, but then the text becomes unreadable.
Is there any way to generate the image without a slanted font?

flodolo:
Maybe something like this?

TheoChevalier:
Comparison with my tool, not a big difference https://lut.im/m8j86BsJpb/kg8LgyGnXQWnJBQG.png

flodolo:
404 alternative image
404cloud

TheoChevalier:
Can you have it in black only? (Or with a transparent background?)

flodolo:
The second one has a transparent background (it's done with the same service you're using).

This one is black with transparent background (personally I like having some variations in colors).

TheoChevalier:
Nice, I was just looking for a transparent background :) I like the one with color variations, and I guess that if we switch the darker blue to #1C75BC (Color of Transvision logo), it would be too close the the lighter blue

flodolo:
This one uses black, #0095dd (links color) and #1C75BC



This one has a slightly dark background (black at 10% opacity)

404_bg

pascalchevrel:
The page and it's different variations look great but it should return a 404 HTTP header, not 200 ;)

flodolo:
It looks great to me now.

@pascalchevrel 
Ready to merge?

PR #717, opened 2016-04-02
Fix broken search results highlighting

### LINKED ISSUE #715: Highlight of match in results does not work properly
I noticed that a perfect match in the results isn't highlighted correct.

See https://transvision.mozfr.org/?recherche=+Add+a+Search+Engine&repo= Add a Search Engine aurora&sourcelocale=en-US&locale=da&search_type=strings

Some of the chars in 'Add a Search Engine' is not highlighted.

flodolo:
Fix #715

### DIFF
--- a/app/classes/Transvision/Utils.php
+++ b/app/classes/Transvision/Utils.php
@@ -119,8 +119,21 @@ class Utils
      */
     public static function highlightString($str)
     {
+        /*
+            [^←→]: matches any character but the marker characters.
+            [^←→]|(?R): matches any character but the marker characters,
+                        or recursively the entire pattern.
+
+            This is wrapped in a non capturing group (?:), since we're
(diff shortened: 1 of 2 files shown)


app/classes/Transvision/Utils.php
```diff
@@ -119,8 +119,9 @@ public static function markString($needle, $haystack)
      */
     public static function highlightString($str)
     {
+        // See http://stackoverflow.com/a/14952740 for the regex background
         $str = preg_replace(
-            '/←(.*)→/isU',
+            '/←((?:[^←→]|(?R))*)→/isu',
```
flodolo: On second thoughts we should be able to drop also the 's' [modifier](http://php.net/manual/en/reference.pcre.pattern.modifiers.php). And probably worth adding a better explanation than a link to SO.

PR #718, opened 2016-04-03
Fix the comment: replace “semi-colon” by “colon”

djug:
The comment is referring to semi-colons while the router is replacing colons


flodolo:
Thanks :+1:

PR #749, opened 2016-05-27
updating php-cs-fixer path

xXGuillaumeXx:


pascalchevrel:
Looks good, I think it may be worth updating to a more recent version too, like 1.9.x and run this on our source code as part of the same PR in case it catches more stuff than the 1.4 version we currently use.

PR #754, opened 2016-06-19
Fixes #743: Add a button to transliterate words in the main search view

### LINKED ISSUE #743: Add a button to transliterate words in the main search view
Using an idea introduced in #699.

We could display a "Transliterate" button for Serbian only that would replace the Cyrillic results with transliterated Latin (by putting both strings in each cell and using display: none to switch between them)

SkySymbol:
This adds a button to toggle between Serbian Cyrillic and Serbian Latin.

### DIFF
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -315,7 +315,7 @@ class ShowResults
             $regular_string_id = 'string_' . $string_id;
 
             // Find if we need to transliterate the string
-            $transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string != '@@missing@@';
+            $transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string && $target_string != '@@missing@@';
 
             if ($transliterate) {
                 $transliterated_string = self::getTransliteratedString($target_string, 'sr-Cyrl');
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -314,7 +314,11 @@ class ShowResults
             $string_id = md5($key . mt_rand());
             $regular_string_id = 'string_' . $string_id;
 
-            // Find if we need to transliterate the string
+            /*
+                Find if we need to transliterate the string.
+                The string gets transliterated if the target local is serbian,
+                if we aren't in the 3locales view and if we have a $target_string
+            */
             $transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string && $target_string != '@@missing@@';
 
             if ($transliterate) {
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -403,8 +403,8 @@ class ShowResults
                 if ($transliterate) {
                     $meta_target .= "<input type='button' value='To Latin' data-transliterated-id='{$string_id}' class='transliterate_button button action'>";
                 }
+                $meta_target .= $error_message;
             }
-            $meta_target .= "{$error_message}";
 
             // If there is no target_string2, display an error, otherwise display the string + meta links
             if ($target_string2 == '@@missing@@') {
@@ -480,9 +480,7 @@ class ShowResults
                       <a class='bug_link' target='_blank' href='{$bz_link[0]}'>
                         &lt;report a bug&gt;
                       </a>
-                      {$meta_target}";
-            $table .= "
-                      {$error_message}
+                      {$meta_target}
                     </div>
                   </td>
                 {$extra_column_rows}
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -145,7 +145,7 @@ class ShowResults
     }
 
     /**
-     * Use the api to return a transliterated string
+     * Use the API to return a transliterated string
      *
      * @param string $string The string to be transliterated
      * @param string $locale The target locale
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -322,7 +322,7 @@ class ShowResults
             $transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string && $target_string != '@@missing@@';
 
             if ($transliterate) {
-                $transliterated_string = self::getTransliteratedString($target_string, 'sr-Cyrl');
+                $transliterated_string = self::getTransliteratedString(urlencode($target_string), 'sr-Cyrl');
                 $transliterate_string_id = 'transliterate_' . $string_id;
             }
 
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -315,9 +315,7 @@ class ShowResults
             $regular_string_id = 'string_' . $string_id;
 
             // Find if we need to transliterate the string
-            $transliterate = ($locale2 == 'sr' && ! $extra_locale && $target_string)
-                ? true
-                : false;
+            $transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string != '@@missing@@';
 
             if ($transliterate) {
                 $transliterated_string = self::getTransliteratedString($target_string, 'sr-Cyrl');
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -145,6 +145,21 @@ class ShowResults
     }
 
     /**
+     * Use the api to return a transliterated string
+     *
+     * @param string $string The string to be transliterated
+     * @param string $locale The target locale
+     *
+     * @return string The string transliterated into the target locale
+     */
+    public static function getTransliteratedString($string, $locale)
+    {
+        $request = new API(parse_url(API_ROOT . "transliterate/$locale/$string"));
+        $json = include MODELS . 'api/transliterate.php';
+
+        return $json[0];
+    }
+    /**
      * Return search results in a repository on strings/entities for the API
      *
      * @param  array $entities      An array of all the entities we want to return
@@ -296,21 +311,43 @@ class ShowResults
                 $target_string2 = '';
             }
 
+            $string_id = md5($target_string);
+            $regular_string_id = 'string_' . $string_id;
+
+            // Find if we need to transliterate the string
+            $transliterate = ($locale2 == 'sr' && ! $extra_locale && $target_string)
+                ? true
+                : false;
+
+            if ($transliterate) {
+                $transliterated_string = self::getTransliteratedString($target_string, 'sr-Cyrl');
+                $transliterate_string_id = 'transliterate_' . $string_id;
+            }
+
             foreach ($search as $val) {
                 $source_string = Utils::markString($val, $source_string);
                 $target_string = Utils::markString($val, $target_string);
                 if ($extra_locale) {
                     $target_string2 = Utils::markString($val, $target_string2);
                 }
+                if ($transliterate) {
+                    $transliterated_string = Utils::markString($val, $transliterated_string);
+                }
             }
 
             // Escape HTML before highlighing search terms
             $source_string = htmlspecialchars($source_string);
             $target_string = htmlspecialchars($target_string);
+
             $source_string = Utils::highlightString($source_string);
             $target_string = Utils::highlightString($target_string);
             $source_string = Strings::highlightSpecial($source_string);
             $target_string = Strings::highlightSpecial($target_string);
+
+            if ($transliterate) {
+                $transliterated_string = htmlspecialchars($transliterated_string);
+                $transliterated_string = Utils::highlightString($transliterated_string);
+            }
 
             if ($extra_locale) {
                 $target_string2 = htmlspecialchars($target_string2);
@@ -362,10 +399,12 @@ class ShowResults
             } elseif (! $target_string) {
                 $target_string = '<em class="error">Warning: Empty string</em>';
             } else {
-                $meta_target = "
-                      <span class='clipboard' data-clipboard-target='#{$clipboard_target_string}' alt='Copy to clipboard'></span>
-                      {$error_message}";
+                $meta_target = "<span class='clipboard' data-clipboard-target='#{$regular_string_id}' alt='Copy to clipboard'></span>";
+                if ($transliterate) {
+                    $meta_target .= "<input type='button' value='To Latin' data-transliterated-id='{$string_id}' class='transliterate_button button action'>";
+                }
             }
+            $meta_target .= "{$error_message}";
 
             // If there is no target_string2, display an error, otherwise display the string + meta links
             if ($target_string2 == '@@missing@@') {
@@ -428,7 +467,11 @@ class ShowResults
 
                   <td dir='{$direction2}' lang='{$locale2}'>
                     <span class='celltitle'>{$locale2}</span>
-                    <div class='string' id='{$clipboard_target_string}'>{$target_string}</div>
+                    <div class='string' id='{$regular_string_id}'>{$target_string}</div>";
+            if ($transliterate) {
+                $table .= "<div class='string toggle' id='{$transliterate_string_id}' style='display: none;'>{$transliterated_string}</div>";
+            }
+            $table .= "
                     <div dir='ltr' class='result_meta_link'>
                       <a class='source_link' href='{$locale2_path}'>
                         &lt;source&gt;
@@ -437,7 +480,9 @@ class ShowResults
                       <a class='bug_link' target='_blank' href='{$bz_link[0]}'>
                         &lt;report a bug&gt;
                       </a>
-                      {$meta_target}
+                      {$meta_target}";
+            $table .= "
+                      {$error_message}
                     </div>
                   </td>
                 {$extra_column_rows}
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -311,7 +311,7 @@ class ShowResults
                 $target_string2 = '';
             }
 
-            $string_id = md5($target_string);
+            $string_id = md5($key . mt_rand());
             $regular_string_id = 'string_' . $string_id;
 
             // Find if we need to transliterate the string
--- a/app/inc/constants.php
+++ b/app/inc/constants.php
@@ -17,6 +17,7 @@ define('CONTROLLERS',   APP_ROOT . 'controllers/');
 define('CACHE_ENABLED', isset($_GET['nocache']) ? false : true);
 define('CACHE_PATH',    INSTALL_ROOT . 'cache/');
 define('APP_SCHEME',    isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off' ? 'https://' : 'http://');
+define('API_ROOT',  APP_SCHEME . htmlspecialchars($_SERVER['HTTP_HOST'], ENT_QUOTES, 'UTF-8') . '/api/v1/');
 
 /*
     Determine the last Git hash from cache/version.txt, ignoring new lines.
--- a/web/js/toggle_transliterated_string.js
+++ b/web/js/toggle_transliterated_string.js
@@ -1,9 +1,9 @@
 $(document).ready(function() {
     $('.transliterate_button').click(function(event) {
-        var $transliterated = $('#transliterate_' + event.target.getAttribute('data-transliterated-id'));
-        var $normal = $('#string_' + event.target.getAttribute('data-transliterated-id'));
-        event.preventDefault();
         var id = event.target.getAttribute('data-transliterated-id');
+        var $transliterated = $('#transliterate_' + id);
+        var $normal = $('#string_' + id);
+        event.preventDefault();
         $normal.toggle();
         $transliterated.toggle();
         if ($normal.css('display') === 'none') {
--- a/web/js/toggle_transliterated_string.js
+++ b/web/js/toggle_transliterated_string.js
@@ -1,0 +1,17 @@
+$(document).ready(function() {
+    $('.transliterate_button').click(function(event) {
+        var $transliterated = $('#transliterate_' + event.target.getAttribute('data-transliterated-id'));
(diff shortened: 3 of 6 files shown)


TheoChevalier:
Thanks for the patch @SkySymbol 
I’ll keep testing in a few days after my exams, but right now it seem to break the main search :)
This is what I get by clicking on the main search field and hitting Search. Results are not returned. capture d ecran 2016-06-19 a 14 56 19
At first glance I would blame the JS file

app/classes/Transvision/ShowResults.php
```diff
@@ -397,6 +412,10 @@ public static function resultsTable($search_object, $search_results, $page)
             $anchor_name = str_replace(['/', ':'], '_', $key);
 
             $clipboard_target_string  = 'clip_' . md5($target_string);
+            if ($locale2 == 'sr') {
+                $transliterated_string = self::getTransliteratedString($target_string, 'sr-Cyrl');
+                $transliterate_id = substr($clipboard_target_string, 5);
```
TheoChevalier: I don’t think we should reuse the clipboard id, maybe something like that?
`$transliterate_id = 'transliterate_' . md5($target_string);`

app/inc/constants.php
```diff
@@ -17,6 +17,7 @@
 define('CACHE_ENABLED', isset($_GET['nocache']) ? false : true);
 define('CACHE_PATH',    INSTALL_ROOT . 'cache/');
 define('APP_SCHEME',    isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off' ? 'https://' : 'http://');
+define('API_ROOT',   'http://' . htmlspecialchars($_SERVER['HTTP_HOST'], ENT_QUOTES, 'UTF-8') . '/api/v1/');
```
TheoChevalier: We could reuse the APP_SCHEME defined above instead of hardcoding http.
Also, I think this won’t work with setups like http://server.com/some/path/transvision/api/v1. Do we really need the host name here?
+ nit: indent

TheoChevalier:
Tried with 3 locales search, the button is not displayed when sr is the source locale or the extra locale. I think we should also hide it when the string is missing. Since there are other info already displayed when a string is missing, I’ll open an issue for that.

app/classes/Transvision/ShowResults.php
SkySymbol: The issue i have is that I have to identify the Cyrillic string and the transliterated one with nearly the same id because the button must be associated with its corresponding texts and I use the id of the button to find the right text to toggle [here](https://github.com/SkySymbol/transvision/blob/button_transliterate/web/js/toggle_transliterated_string.js#L3-4). And the ```<div>``` for the target string is already identified with ```$clipboard_id```.

I don't know if it's the right approach.

TheoChevalier: Ok, got it. 
To make things less confusing, I suggest we rename ids as following:
string_ + md5 for the regular string (instead of clip_, even for the clipboard feature)
transliterated_ + md5 for the transliterated string. 

Then you can use `$transliterate_id = 'transliterate_' . md5($target_string);` instead of the substr. For the button we can then replace the  id with a custom attribute (data-transliterated-id?) since we’re not using it as a real id but only to store data to build the actual id with JS. 

Also, we should be able to copy the Latin string with the clipboard button when the transliterated string is displayed. To do that, when we toggle divs we would need to update the data-clipboard-target attribute on the clipboard span with the right id (string_ + md5 or transliterated_ + md5)

app/classes/Transvision/ShowResults.php
```diff
@@ -456,15 +475,23 @@ public static function resultsTable($search_object, $search_results, $page)
...
                       <a class='source_link' href='{$locale2_path}'>
                         &lt;source&gt;
                       </a>
                       &nbsp;
                       <a class='bug_link' target='_blank' href='{$bz_link[0]}'>
                         &lt;report a bug&gt;
-                      </a>
+                      </a>";
+            if ($locale2 == 'sr') {
+                $table .= "<input type='button' value='To latin' id='{$transliterate_id}'class='transliterate_button'>";
```
TheoChevalier: Missing space before class, making html invalid

web/js/toggle_transliterated_string.js
```diff
@@ -0,0 +1,14 @@
+$(document).ready(function() {
+    $('input[value="To latin"],input[value="To cyrillic"]').click(function(event) {
```
TheoChevalier: You can use `$('.transliterate_button').click(function(event) {`

web/js/toggle_transliterated_string.js
```diff
@@ -0,0 +1,14 @@
...
+        var $transliterated = $('#string_' + event.target.id);
+        var $normal = $('#clip_' + event.target.id);
+        event.preventDefault();
+        $transliterated.toggle();
+        $normal.toggle();
+        if ($normal.css('display') === 'none') {
+            $(this).val('To cyrillic');
```
TheoChevalier: nit: Cyrillic

web/js/toggle_transliterated_string.js
```diff
@@ -0,0 +1,14 @@
...
+        } else {
+            $(this).val('To latin');
```
TheoChevalier: nit: Latin

app/inc/constants.php
```diff
@@ -42,6 +42,22 @@
     define('CACHE_TIME',  19200);
 }
 
+$api_root = APP_SCHEME . htmlspecialchars($_SERVER['HTTP_HOST'], ENT_QUOTES, 'UTF-8');
+
+// trying to find the root path of transvision
+$root_path = explode('transvision', htmlspecialchars($_SERVER['REQUEST_URI'], ENT_QUOTES, 'UTF-8'));
```
TheoChevalier: My bad, I think in my previous comment I was confused with other projects were the webroot can be anything. On Transvision, we actually can’t have anything like `https://domain.tld/foo/bar/api/v1`. It will always be `https://domain.tld/api/v1/`.
So this should be enough:
`APP_SCHEME . htmlspecialchars($_SERVER['HTTP_HOST'], ENT_QUOTES, 'UTF-8') . '/api/v1/';` (probably what you did in the previous version)
However, HTTP_HOST is not necessarily defined, so we should at least check it’s defined. Not sure if gethostname would work with php integrated server, but might be worth trying https://secure.php.net/manual/en/function.gethostname.php

TheoChevalier:
This query triggers a warning: http://localhost:8082/3locales/?recherche=test&repo=aurora&sourcelocale=en-US&locale=sr&locale2=sr&search_type=strings

`PHP Notice:  Undefined offset: 3 in /home/theo/moz/transvision/app/models/api/transliterate.php on line 24`

app/classes/Transvision/ShowResults.php
```diff
@@ -461,7 +487,13 @@ public static function resultsTable($search_object, $search_results, $page)
                       <a class='bug_link' target='_blank' href='{$bz_link[0]}'>
                         &lt;report a bug&gt;
                       </a>
-                      {$meta_target}
+                      {$meta_target}";
+            if ($locale2 == 'sr' && !$extra_locale) {
```
TheoChevalier: You need to move this block earlier, around line 400

app/classes/Transvision/ShowResults.php
```diff
@@ -461,7 +487,13 @@ public static function resultsTable($search_object, $search_results, $page)
...
+                $table .= "<input type='button' value='To latin' data-transliterated-id='{$string_id}' class='transliterate_button'>";
+            }
+            $table .= "
+                      <span class='clipboard' data-clipboard-target='#{$regular_string_id}' alt='Copy to clipboard'><img src='/img/copy_icon_black_18x18.png'></span>
```
TheoChevalier: This should be removed, it’s already here https://github.com/mozfr/transvision/pull/754/files#diff-4d976c6b5c0ea357b6a48453ca776e91R405

app/classes/Transvision/ShowResults.php
TheoChevalier: also, missing space `! $extra_locale`

app/classes/Transvision/ShowResults.php
```diff
@@ -452,7 +474,11 @@ public static function resultsTable($search_object, $search_results, $page)
 
                   <td dir='{$direction2}' lang='{$locale2}'>
                     <span class='celltitle'>{$locale2}</span>
-                    <div class='string' id='{$clipboard_target_string}'>{$target_string}</div>
+                    <div class='string' id='{$regular_string_id}'>{$target_string}</div>";
+            if ($locale2 == 'sr' && !$extra_locale) {
```
TheoChevalier: same here, `! $extra_locale`

app/classes/Transvision/ShowResults.php
```diff
@@ -461,7 +487,13 @@ public static function resultsTable($search_object, $search_results, $page)
                       <a class='bug_link' target='_blank' href='{$bz_link[0]}'>
                         &lt;report a bug&gt;
                       </a>
-                      {$meta_target}
+                      {$meta_target}";
+            if ($locale2 == 'sr' && !$extra_locale) {
+                $table .= "<input type='button' value='To latin' data-transliterated-id='{$string_id}' class='transliterate_button'>";
```
TheoChevalier: To Latin

app/classes/Transvision/ShowResults.php
```diff
@@ -403,6 +418,13 @@ public static function resultsTable($search_object, $search_results, $page)
             // Replace / and : in the key name and use it as an anchor name
             $anchor_name = str_replace(['/', ':'], '_', $key);
 
+            $string_id = md5($target_string);
+            $regular_string_id = 'string_' . $string_id;
+            if ($locale2 == 'sr' && !$extra_locale) {
```
TheoChevalier: and here, too `! $extra_locale`

app/classes/Transvision/ShowResults.php
```diff
@@ -403,6 +418,13 @@ public static function resultsTable($search_object, $search_results, $page)
...
+                $transliterated_string = self::getTransliteratedString($target_string, 'sr-Cyrl');
```
TheoChevalier: we actually need to do this call right after we define $target_string, and before we do all the transforms  (highlight, markString, htmlspecialchars…), otherwise tags are displayed in the transliterated string. We will also need to apply the same transforms on our transliterated string.
I think everything is here https://github.com/mozfr/transvision/blob/master/app/classes/Transvision/ShowResults.php#L318-335

TheoChevalier:
I think we should remove the `if ($locale2 == 'sr' && ! $extra_locale) {` checks and replace them with a single Boolean ($transliterate ?), especially since we will need more of them to apply transforms.

SkySymbol:
Thank you for taking the time to review my pull request!

TheoChevalier:
Thanks for the updates :) Before I forget, you don’t need to squash your commits anymore, GitHub has a new feature that lets us do it automatically before merging. This will probably make your life easier and also helps reading changes after each update

TheoChevalier:
With the following query, I got a `Notice: Undefined offset: 3 in /Users/tchevalier/moz/transvision/app/models/api/transliterate.php on line 24` For an unknown reason, this string is considered empty when it’s passed to the transliterate function.

http://localhost:8082/?recherche=%231+%232+%D0%BD%D0%B8%D1%98%D0%B5+%D0%BA%D0%BE%D0%BC%D0%BF%D0%B0%D1%82%D0%B8%D0%B1%D0%B8%D0%BB%D0%B0%D0%BD+%D1%81%D0%B0+%D0%BD%D0%B0%D1%98%D0%BD%D0%BE%D0%B2%D0%B8%D1%98%D0%BE%D0%BC+%D0%B2%D0%B5%D1%80%D0%B7%D0%B8%D1%98%D0%BE%D0%BC+Firefox+Sync-%D0%B0.+%D0%90%D0%B6%D1%83%D1%80%D0%B8%D1%80%D0%B0%D1%98%D1%82%D0%B5+%D0%B4%D0%BE+%D0%BD%D0%B0%D1%98%D0%BD%D0%BE%D0%B2%D0%B8%D1%98%D0%B5+%D0%B2%D0%B5%D1%80%D0%B7%D0%B8%D1%98%D0%B5.&repo=aurora&sourcelocale=en-US&locale=sr&search_type=strings&perfect_match=perfect_match

app/classes/Transvision/ShowResults.php
```diff
@@ -315,20 +330,42 @@ public static function resultsTable($search_object, $search_results, $page)
                 $target_string2 = '';
             }
 
+            $string_id = md5($target_string);
+            $regular_string_id = 'string_' . $string_id;
+
+            // Find if we need to transliterate the string
+            $transliterate = ($locale2 == 'sr' && ! $extra_locale && $target_string)
```
TheoChevalier: You can use `$transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string != '@@missing@@';` no need to specify true when true and false when false ;)

app/classes/Transvision/ShowResults.php
```diff
@@ -386,10 +423,12 @@ public static function resultsTable($search_object, $search_results, $page)
             } elseif (! $target_string) {
                 $target_string = '<em class="error">Warning: Empty string</em>';
             } else {
-                $meta_target = "
-                      <span class='clipboard' data-clipboard-target='#{$clipboard_target_string}' alt='Copy to clipboard'><img src='/img/copy_icon_black_18x18.png'></span>
-                      {$error_message}";
+                $meta_target = "<span class='clipboard' data-clipboard-target='#{$regular_string_id}' alt='Copy to clipboard'><img src='/img/copy_icon_black_18x18.png'></span>";
+                if ($transliterate) {
+                    $meta_target .= "<input type='button' value='To latin' data-transliterated-id='{$string_id}' class='transliterate_button'>";
```
TheoChevalier: To Latin ;)

app/classes/Transvision/ShowResults.php
```diff
@@ -461,7 +504,9 @@ public static function resultsTable($search_object, $search_results, $page)
                       <a class='bug_link' target='_blank' href='{$bz_link[0]}'>
                         &lt;report a bug&gt;
                       </a>
-                      {$meta_target}
+                      {$meta_target}";
```
TheoChevalier: You forgot to remove this part, only {$meta_target} should remain here, since we’re defining its content earlier

app/classes/Transvision/ShowResults.php
```diff
@@ -386,10 +423,12 @@ public static function resultsTable($search_object, $search_results, $page)
...
             } else {
-                $meta_target = "
-                      <span class='clipboard' data-clipboard-target='#{$clipboard_target_string}' alt='Copy to clipboard'><img src='/img/copy_icon_black_18x18.png'></span>
-                      {$error_message}";
+                $meta_target = "<span class='clipboard' data-clipboard-target='#{$regular_string_id}' alt='Copy to clipboard'><img src='/img/copy_icon_black_18x18.png'></span>";
+                if ($transliterate) {
+                    $meta_target .= "<input type='button' value='To latin' data-transliterated-id='{$string_id}' class='transliterate_button'>";
+                }
             }
+            $meta_target .= "{$error_message}";
```
TheoChevalier: This line should be inside the else statement above, since we want to display the error only when the string is not empty. Since it’s only concatenating a variable, it should be `$meta_target .= $error_message;`

TheoChevalier:
I think we’ll have to think of a better string id since the same $target_string can be displayed several times on the same page (thus generating the same md5 hash, and making several divs toggle). I will think about it when a rested brain :)

SkySymbol:
Could we use ``$entity_link`` and ``$target_string`` to make unique hashes?

TheoChevalier:
No, that would not be unique. Look at this query for instance: https://transvision.mozfr.org/?recherche=Firefox&repo=release&sourcelocale=en-US&locale=fr&search_type=strings&perfect_match=perfect_match

This string would have the same ID in both result tables browser/branding/official/brand.properties:brandShortName
This would work in theory: $key + [locale of the result table], but we don’t know the later in resultsTable().

TheoChevalier:
I suggest we go for `md5($key . mt_rand())`, there’s nothing differentiating the two lines in the example above and I don’t think we should add an argument to the function only for that.

app/classes/Transvision/ShowResults.php
```diff
@@ -145,6 +145,21 @@ public static function getTranslationMemoryResults($strings, $search, $max_resul
     }
 
     /**
+     * Use the api to return a transliterated string
```
flodolo: nit: API

app/classes/Transvision/ShowResults.php
```diff
@@ -315,20 +330,40 @@ public static function resultsTable($search_object, $search_results, $page)
                 $target_string2 = '';
             }
 
+            $string_id = md5($key . mt_rand());
+            $regular_string_id = 'string_' . $string_id;
+
+            // Find if we need to transliterate the string
```
flodolo: It would be nice to explain the logic of the following expression in the comment

web/js/toggle_transliterated_string.js
```diff
@@ -0,0 +1,17 @@
+$(document).ready(function() {
+    $('.transliterate_button').click(function(event) {
+        var $transliterated = $('#transliterate_' + event.target.getAttribute('data-transliterated-id'));
+        var $normal = $('#string_' + event.target.getAttribute('data-transliterated-id'));
+        event.preventDefault();
+        var id = event.target.getAttribute('data-transliterated-id');
```
flodolo: Let's move this at the beginning and use the variable instead of calling getAttribute 3 times

app/classes/Transvision/ShowResults.php
```diff
@@ -145,6 +145,21 @@ public static function getTranslationMemoryResults($strings, $search, $max_resul
...
+     * @param string $locale The target locale
+     *
+     * @return string The string transliterated into the target locale
+     */
+    public static function getTransliteratedString($string, $locale)
+    {
+        $request = new API(parse_url(API_ROOT . "transliterate/$locale/$string"));
+        $json = include MODELS . 'api/transliterate.php';
+
+        return $json[0];
```
flodolo: This line generates a ton of errors in my VM.
```
 Notice: Undefined offset: 0 in /home/flodolo/github/transvision/app/classes/Transvision/ShowResults.php on line 160
```

app/classes/Transvision/ShowResults.php
```diff
@@ -452,7 +489,11 @@ public static function resultsTable($search_object, $search_results, $page)
 
                   <td dir='{$direction2}' lang='{$locale2}'>
                     <span class='celltitle'>{$locale2}</span>
-                    <div class='string' id='{$clipboard_target_string}'>{$target_string}</div>
+                    <div class='string' id='{$regular_string_id}'>{$target_string}</div>";
+            if ($transliterate) {
+                $table .= "<div class='string toggle' id='{$transliterate_string_id}' style='display: none;'>{$transliterated_string}</div>";
```
flodolo: Are you sure you need the style here?

app/classes/Transvision/ShowResults.php
flodolo: OK, my VM was missing php5-intl. Still, I think we should be a little more error proof.

flodolo:
One bug: it messes up strings with HTML code.

For example, search for ` Delete Recent <html:strong>Cookies</html:strong>`

Button is also not showing up on entities searches

TheoChevalier:
> Button is also not showing up on entities searches

Yeah, was planning on back-porting the code as a follow up once this PR is finalized (but of course, it can be done here)

SkySymbol:
> With the following query, I got a Notice: Undefined offset: 3 in /Users/tchevalier/moz/transvision/app/models/api/transliterate.php on line 24 For an unknown reason, this string is considered empty when it’s passed to the transliterate function.

> http://localhost:8082/?recherche=%231+%232+%D0%BD%D0%B8%D1%98%D0%B5+%D0%BA%D0%BE%D0%BC%D0%BF%D0%B0%D1%82%D0%B8%D0%B1%D0%B8%D0%BB%D0%B0%D0%BD+%D1%81%D0%B0+%D0%BD%D0%B0%D1%98%D0%BD%D0%BE%D0%B2%D0%B8%D1%98%D0%BE%D0%BC+%D0%B2%D0%B5%D1%80%D0%B7%D0%B8%D1%98%D0%BE%D0%BC+Firefox+Sync-%D0%B0.+%D0%90%D0%B6%D1%83%D1%80%D0%B8%D1%80%D0%B0%D1%98%D1%82%D0%B5+%D0%B4%D0%BE+%D0%BD%D0%B0%D1%98%D0%BD%D0%BE%D0%B2%D0%B8%D1%98%D0%B5+%D0%B2%D0%B5%D1%80%D0%B7%D0%B8%D1%98%D0%B5.&repo=aurora&sourcelocale=en-US&locale=sr&search_type=strings&perfect_match=perfect_match

I've found the issue but I'm not sure about how to implement the fix. The issue comes from here https://github.com/mozfr/transvision/blob/master/app/classes/Transvision/API.php#L74. The parameter is not passed to the api because `$url['path']` does not contain it, because `parse_url` does not include it, because of the leading '#'. See http://php.net/manual/en/function.parse-url.php#refsect1-function.parse-url-returnvalues for more details.

The fix would be to add `$url['fragment']` to the parameters to `getParameters` if it is defined

TheoChevalier:
Good catch, I think you nailed it. I think I’ve found an easy way to fix it, using urldecode and urlencode https://pastebin.mozilla.org/8888285 It seem to work

flodolo:
Failure on Travis will go away if you rebase to master.

TheoChevalier:
Hm, looks like I’ve introduced a conflict with the hi-DPI PR… There will be another conflict after #787 is merged, but I think I’ll be done after that. Sorry for the troubles :/

I haven’t tested the latest changes here, but I think we’re almost ready :)

SkySymbol:
Having conflicts is a good thing because it means that the project is moving :)

TheoChevalier:
I squashed your commits into a single one and added changes on top of it (adding the button to entity view). Let’s close this PR and move to #792

PR #757, opened 2016-07-02
Remove "buttons on hover" experiment, fixes #756

### LINKED ISSUE #756: Consider reverting the "buttons on hover" change
Should we revert the experiment landed in #728 ?
After a few weeks of using it, I find the behavior more misleading than actually improving the UX. For instance, when a single result is displayed on a big screen, the cursor is likely not to be over the result row and it’s not obvious you have to move it over.
This would also fix the layout issue we noticed when string ids are displayed in multiple lines (weird indents)

TheoChevalier:


flodolo:
Looks definitely good for me, let's see if @pascalchevrel has objections before merging.

pascalchevrel:
you use it more than me so if you think it's better without, let's got this route.

flodolo:
Thanks, merging then.

PR #760, opened 2016-07-04
Actually display abnormal string length errors

TheoChevalier:
Haven’t seen a “Large string?” error in a while? That’s because we were displaying them only on 3locales view :)

### DIFF
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -372,7 +372,7 @@ class ShowResults
                         $error_message = $error_message . '<em class="error"> Small string?</em>';
                         break;
                     case 'large':
-                        $error_message = $error_message . '<em class="error"> Large String?</em>';
+                        $error_message = $error_message . '<em class="error"> Large string?</em>';
                         break;
                 }
             }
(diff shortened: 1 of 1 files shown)


flodolo:
Uh, I never saw those. Do you have an example where it shows up?

Can we also fix the inconsistent case here
https://github.com/mozfr/transvision/pull/760/files#diff-4d976c6b5c0ea357b6a48453ca776e91R375

TheoChevalier:
Yeah, you should see them with those queries:

/?recherche=%231+will+be+installed+after+you+restart+%233.%3B%232+add-ons+will+be+installed+after+you+restart+%233.&repo=aurora&sourcelocale=en-US&locale=ga-IE&search_type=strings

/?recherche=Caution%3A+This+site+would+like+to+install+an+unverified+add-on+in&repo=aurora&sourcelocale=en-US&locale=ga-IE&search_type=strings

Also just noticed it’s not displayed in entity search, I should fix that as well

TheoChevalier:
It made actually no sense to hide error for target_string whether target_string2 was set or not. So I removed this check. This code does not exist anymore in issue #762, so I fixed the issue there and will close this one.

Issue #755, opened 2016-06-19
Hide clipboard button when string is missing


TheoChevalier:
Currently we’re displaying the clipboard button even when the string is missing, I think we should hide it in this case.
I think we should keep other links, though (file a bug, source) as people might want to report the missing string or check where it should be added.

TheoChevalier:
Fixed with 043455deee0980d82c2ae5b5fd90baa64f2deb32

PR #762, opened 2016-07-04
Display the same info in each mainsearch search type

TheoChevalier:
Displaying the same info and features on string, entities and string&entities search types. Also displaying the <source> link in yellow everywhere for consistency.

This add "Translate with", the clipboard button and abnormal string length errors to the entity search. (Follow up to issue #760)

### DIFF
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -340,6 +340,7 @@ class ShowResults
             ];
 
             $target_string = Strings::multipleStringReplace($replacements, $target_string);
+            $clipboard_target_string  = 'clip_' . md5($target_string);
 
             $temp = explode('-', $locale1);
             $locale1_short_code = $temp[0];
@@ -363,7 +364,7 @@ class ShowResults
 
             // If there is no source_string, display an error, otherwise display the string + meta links
             if (! $source_string) {
-                $source_string = '<em class="error">warning: Source string is empty</em>';
+                $source_string = '<em class="error">Warning: Source string is empty</em>';
             } else {
                 $meta_source = "
                     <div dir='ltr' class='result_meta_link'>
@@ -383,7 +384,7 @@ class ShowResults
 
             // If there is no target_string, display an error, otherwise display the string + meta links
             if (! $target_string) {
-                $target_string = '<em class="error">warning: missing string</em>';
+                $target_string = '<em class="error">Warning: Missing string</em>';
             } else {
                 $meta_target = "
                     <div dir='ltr' class='result_meta_link'>
@@ -401,7 +402,7 @@ class ShowResults
 
             // If there is no target_string2, display an error, otherwise display the string + meta links
             if (! $target_string2) {
-                $target_string2 = '<em class="error">warning: missing string</em>';
+                $target_string2 = '<em class="error">Warning: Missing string</em>';
             } else {
                 $meta_target2 = "
                     <div dir='ltr' class='result_meta_link'>
@@ -418,8 +419,6 @@ class ShowResults
 
             // Replace / and : in the key name and use it as an anchor name
             $anchor_name = str_replace(['/', ':'], '_', $key);
-
-            $clipboard_target_string  = 'clip_' . md5($target_string);
 
             // 3locales view
             if ($extra_locale) {
@@ -518,7 +517,7 @@ class ShowResults
         // Check for final dot
         if (substr(strip_tags($source_string), -1) == '.'
             && substr(strip_tags($target_string), -1) != '.') {
-            $error_message = '<em class="error"> No final dot?</em>';
+            $error_message = '<em class="error">No final dot?</em> ';
         }
 
         // Check abnormal string length
@@ -526,10 +525,10 @@ class ShowResults
         if ($length_diff) {
             switch ($length_diff) {
                 case 'small':
-                    $error_message = $error_message . '<em class="error"> Small string?</em>';
+                    $error_message = $error_message . '<em class="error">Small string?</em> ';
                     break;
                 case 'large':
-                    $error_message = $error_message . '<em class="error"> Large string?</em>';
+                    $error_message = $error_message . '<em class="error">Large string?</em> ';
                     break;
             }
         }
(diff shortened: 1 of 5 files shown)


flodolo:
A couple of thoughts: since we're basically duplicating the code, what about moving it to a class function?

Am I the only one who finds very confusing a "missing string" warning in en-US when the string is empty? I think something like "Source string is empty" would make more sense.

I never checked, but I wonder if we can distinguish between empty (on purpose) and missing strings.

TheoChevalier:
Thanks for the feedback ;) That all make sense to me, will look into that

TheoChevalier:
This also fixes issue #755

TheoChevalier:
Only had a quick look, but I think we only have one key: https://github.com/mozfr/transvision/blob/master/app/classes/Transvision/ShowResults.php#L281 So that we can’t tell if the key in the locale’s cache file is present or not. And loading the cache just for that seem not a good idea. So far, that’s the only approach I see to tell if a string is empty or missing

app/classes/Transvision/ShowResults.php
```diff
@@ -355,42 +355,65 @@ public static function resultsTable($search_object, $search_results, $page)
...
-            }
+            // Don't show meta links by default
+            $meta_source = $meta_target = $meta_target2 = '';
 
-            // Missing string error
+            // If there is no source_string, display an error, otherwise display the string + meta links
             if (! $source_string) {
-                $source_string = '<em class="error">warning: missing string</em>';
-                $error_message = '';
+                $source_string = '<em class="error">warning: Source string is empty</em>';
```
flodolo: nit: Warning

app/classes/Transvision/ShowResults.php
```diff
@@ -355,42 +355,65 @@ public static function resultsTable($search_object, $search_results, $page)
...
+                      . "' target='_blank'>Google</a>
+                      <a href='http://www.bing.com/translator/?from={$locale1_short_code}&to={$locale2_short_code}&text="
+                      . urlencode(strip_tags(html_entity_decode(html_entity_decode($source_string))))
+                      . "' target='_blank'>BING</a>
+                    </div>";
             }
 
+            // If there is no target_string, display an error, otherwise display the string + meta links
             if (! $target_string) {
                 $target_string = '<em class="error">warning: missing string</em>';
```
flodolo: nit: Warning, Missing

app/classes/Transvision/ShowResults.php
```diff
@@ -508,4 +500,45 @@ public static function searchEntities($source_strings, $regex)
...
+     *                there is no error.
+     */
+    public static function buildErrorString($source_string, $target_string)
+    {
+        $error_message = '';
+
+        // Check for final dot
+        if (substr(strip_tags($source_string), -1) == '.'
+            && substr(strip_tags($target_string), -1) != '.') {
+            $error_message = '<em class="error"> No final dot?</em>';
```
flodolo: I would really put the space after the `</em>` in all messages, not at the beginning of the label.

app/views/results_entities.php
```diff
@@ -61,19 +61,37 @@
                                               $bz_target_string2, $current_repo, $entity_link)
                   . '">&lt;report a bug&gt;</a>';
 
+        // Don't show meta links by default
+        $meta_target2 = '';
+
+        // If there is no target_string2, display an error, otherwise display the string + meta links
+        if (! $target_string2) {
+            $target_string2 = '<em class="error">warning: missing string</em>';
```
flodolo: nit: Warning, Missing

app/views/results_entities.php
```diff
@@ -85,6 +103,45 @@
...
+
+    // Get the potential errors for $target_string (final dot, long/small string)
+    $error_message = ShowResults::buildErrorString($source_string, $target_string);
+
+    // Don't show meta links by default
+    $meta_source = $meta_target = '';
+
+    // If there is no source_string, display an error, otherwise display the string + meta links
+    if (! $source_string) {
+        $source_string = '<em class="error">warning: source string is empty</em>';
```
flodolo: nit: Warning, Source

app/views/results_entities.php
```diff
@@ -85,6 +103,45 @@
...
+            . "' target='_blank'>Google</a>
+            <a href='http://www.bing.com/translator/?from={$locale1_short_code}&to={$locale2_short_code}&text="
+            . urlencode(strip_tags(html_entity_decode(html_entity_decode($source_string))))
+            . "' target='_blank'>BING</a>
+          </div>";
+    }
+
+    // If there is no target_string, display an error, otherwise display the string + meta links
+    if (! $target_string) {
+        $target_string = '<em class="error">warning: missing string</em>';
```
flodolo: Warning, Missing

flodolo:
I'm getting warnings when searching for Strings and Strings and entities.
```
[Wed Apr 06 07:43:17.687387 2016] [:error] [pid 15933] [client 192.168.133.50:58748] PHP Notice:  Und
efined variable: clipboard_target_string in /home/flodolo/github/transvision/app/classes/Transvision/
ShowResults.php on line 397, referer: http://192.168.133.211/
```

flodolo:
Let's leave the empty string out of this PR. I think we should be smarter in ShowResults::getStringFromEntity, but it doesn't make sense to think about it now.

flodolo:
Code-wise it looks good to me, but after trying out the patch a bit I think there's value in keeping some of the meta links even when the string is empty or missing.

Example: https://transvision.mozfr.org/?recherche=.end&repo=aurora&sourcelocale=en-US&locale=it&search_type=strings_entities

I think the <source> link is useful for both (the file should be there), as it is Report a bug for target (think of a string that is empty but shouldn't, no easy way to report that bug). The only link that should not be displayed is the "Translate with" on an empty source string.

TheoChevalier:
(Not sure why eslint is failing, haven’t touched any JS file in this last commit…)

flodolo:
Looks great, I found another bug though: check for example `/string/?entity=browser/chrome/browser/aboutDialog.dtd:community.exp.start&repo=central`. RTL locales never have the missing string warning, also in the standard search.

I also wonder if we can get rid of the 6px padding on errors.

But let's look at that in a different PR.

Issue #647, opened 2016-02-24
Bug in "Perfect match" search with special characters


flodolo:
https://transvision.mozfr.org/?recherche=...&repo=aurora&sourcelocale=en-US&locale=it&search_type=strings&perfect_match=perfect_match

Searching for '...' gets my all 3 letters strings.

PR #764, opened 2016-07-06
Entities should link to a single string

### LINKED ISSUE #763: Entity link in the result list is ambiguous
How to reproduce:
1. Search e.g. https://transvision.mozfr.org/?sourcelocale=en-US&locale=cs&repo=aurora&search_type=strings&recherche=Not+specified&perfect_match=perfect_match
2. Click an entity link e.g. the second from the bottom https://transvision.mozfr.org/?sourcelocale=en-US&locale=cs&repo=aurora&search_type=entities&recherche=calendar/chrome/calendar/calendar.dtd:priority.level.none
3. There are more results show, extra .accesskey string in this case. But can be more.

IMO this can be fixed by adding `perfect_match=perfect_match` to the link.

TheoChevalier:
Fixes #763, thanks Michal :) Adding `perfect_match=perfect_match` makes sense to me. Maybe it’s been missing for a reason I ignore, but so far I can’t see any.

Branch based on top of #762 since it touches the same code

### DIFF
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -340,6 +340,7 @@ class ShowResults
             ];
 
             $target_string = Strings::multipleStringReplace($replacements, $target_string);
+            $clipboard_target_string  = 'clip_' . md5($target_string);
 
             $temp = explode('-', $locale1);
             $locale1_short_code = $temp[0];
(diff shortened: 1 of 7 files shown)


flodolo:
One possible reason was #647, entities often contain dots.

PR #768, opened 2016-07-06
Differentiate empty from missing strings

### LINKED ISSUE #767: Differentiate empty from missing strings
Right now we display a "Missing string" warning when the string is just empty.

flodolo:
Fix #767

### DIFF
--- a/app/views/onestring.php
+++ b/app/views/onestring.php
@@ -34,7 +34,7 @@ include VIEWS . 'templates/api_promotion.php';
              "  <th><a href='#{$locale}'>{$locale}</a></th>\n";
 
         if (! $translation) {
-            echo "  <td><em class='error'>Warning: Missing string</em></td><td></td>\n";
+            echo "  <td><em class='error'>Warning: Empty string</em></td><td></td>\n";
         } else {
             echo "  <td lang='#{$locale}' {$rtl_support} >" . Utils::secureText($translation) . "</td>\n" .
                  "  <td><a class='onestring_search' href='{$search_link}' title='Search for the entity in this locale'>🔍</a></td>\n";
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -162,7 +162,7 @@ class ShowResults
 
         foreach ($search_results as $entity => $set) {
             // We only want results for which we have a translation
-            if ($set[1]) {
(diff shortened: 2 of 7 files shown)


flodolo:
This also changes the behavior of the onestring view/API request. Right now we're showing an empty string
https://transvision.flod.org/api/v1/entity/central/?id=dom/chrome/dom/dom.properties:GenericImageNamePNG

With this we display only the locales we actually have, which seems more logical.

app/views/onestring.php
```diff
@@ -34,7 +34,7 @@
              "  <th><a href='#{$locale}'>{$locale}</a></th>\n";
 
         if (! $translation) {
-            echo "  <td><em class='error'>Warning: Missing string</em></td><td></td>\n";
+            echo "  <td><em class='error'>Warning: Empty string</em></td><td></td>\n";
```
TheoChevalier: I’m confused why you left only “Empty”. Now on http://localhost:8082/string/?entity=dom/chrome/dom/dom.properties:GenericImageNamePNG&repo=central it’s displaying "Empty string” everywhere while the string is missing.

flodolo:
There something wrong with my local VM, I switched the dev server to my branch and I see your same errors.

Will need to look into it next week.

flodolo:
You're right about showing "Missing string" in empty, I can't understand why the online servers are still showing Venkman, when they shouldn't
https://github.com/mozfr/transvision/pull/709

flodolo:
Let's close this one. I realized it needs a lot more work, starting from extraction :-\

flodolo:
This might be ready for some more tests. To test, you need to recreate TMX caches, you can do that setting `forceTMX` to true in glossaire.sh
https://github.com/mozfr/transvision/blob/master/app/scripts/glossaire.sh#L111

One string (view and API): it should only show the strings we actually have, or strings that are empty.

I tried to go through the code and identify places where we assumed empty=missing, but I only found place where we wrongly assumed that missing strings would be missing from the TMX ([example](https://github.com/mozfr/transvision/blob/master/app/classes/Transvision/Health.php#L63)).

Changes to tmx_products.py are also in its own repo upstream, with tests
https://github.com/flodolo/tmx_maker/pull/4

TheoChevalier:
Looks good 👍 We’ll have to remember recreating TMX files on the server for the next release

PR #769, opened 2016-07-08
Transvision visual refresh

TheoChevalier:
I’ve been working on this last week, I think it’s in a good shape to start gathering feedback.
You can play with it here: https://dev.theochevalier.fr/

It’s a small visual update to refresh the look & feel until we rethink the UI (eventually).
I don’t expect these changes to be disruptive for our users, but of course you may think differently. I could go further with the UI changes, but one step at a time.

To sum up the current changes:

- Design a bit more flat for consistency with the logo (and we were almost there anyway)
- Reducing the number of colors we’re using for consistency
- Switching from Open Sans to Fira Sans, and generally reducing font-weight
- Start moving away from our default "everything is centered" design (still thinking about what else can be done)
- Block the menu button on the right for large screens (got feedback from delphine that it’s quite hard to notice)
- Many other small details to make the UI less clumsy

Before I forget, commits 1 and 4 need to be removed before merging (webhook changes to keep the demo server up-to-date)

@flodolo @pascalchevrel @SkySymbol and others, see any show-stopper? Like it/Don’t like it?

Note: quite broken with IE but I don’t think we should care (and works well with Edge anyway)


MatonAnthony:
I find it more readable ;)

pascalchevrel:
Lots of good things in it. The things I don't like are:
1/ Fira sans, that font is too bold and blurry by default for the main text but Fira Sans Light is fine, what would you think of using Fira Sans Light instead of Fira Sans? Do we have the same language coverage in Fira Sans than in Open Sans? Testing Fulah, Arabic, Polish, Turkish, and CJK would be a good start.

2/ A smaller logo is OK but its placement on the left instead of centered looks a bit off to me

TheoChevalier:
Thanks for the feedback ;)

1/ I agree with Fira Sans regular being too bold. I didn’t dare trying because I thought it would be too thin, but Fira Sans Light looks indeed quite nice with main content. Switching to it and dropping Fira Sans regular (page will be faster to load).
According to Font Squirrel languages section, Fira Sans supports the same languages as Open Sans + Míkmaq, Sami (Northern), Ulithian, so I think we got covered. And this was a release from 2013, we’re using the latest.

The only issue I see is Romanian diacritics but I think it was already broken. For the same string sometimes they are displayed properly, sometimes not. IIRC, mozilla.org got that fixed with a specific version of Open Sans (or hacked it?)
https://transvision.mozfr.org/?sourcelocale=en-US&locale=ro&repo=mozilla_org&search_type=entities&recherche=mozilla_org/firefox/accounts.lang:315b9ec0
https://dev.theochevalier.fr/?sourcelocale=en-US&locale=ro&repo=mozilla_org&search_type=entities&recherche=mozilla_org/firefox/accounts.lang:315b9ec0

2/ I also agree with that and will revert the change unless I can propose something with more left-aligned elements.

SkySymbol:
Great changes! That looks more modern! Maybe we could have a two columns layout for the search and filter options?

TheoChevalier:
Could be something to try, yeah.

Also, just checked our stats, and as surprising as it could be, IE still represent 5% of our traffic in 2016… And that’s excluding Edge, of course (~1%).
Might be people checking Transvision at work 😐

flodolo:
@TheoChevalier Can you rebase the branch?

web/github_hook.php
```diff
@@ -8,7 +8,7 @@
 $composer = $app_root . '/composer.phar';
 
 // Git variables
-$branch = 'master';
+$branch = 'redesign';
```
flodolo: Remember to exclude changes to this file later.

web/style/transvision.css
```diff
@@ -57,20 +41,24 @@ html {
...
     position: relative;
-    background-color: #cadded;
-    background-image: url(/img/grain.png), linear-gradient(to bottom, #cadded 0%, #f6f4ec 100%);
-    background-origin: padding-box, padding-box;
-    background-position: 0 0, 0 0, 0 0;
-    background-repeat: repeat;
-    background-size: auto auto, auto auto, auto auto;
     min-width: 320px;
+    display: flex;
+    min-height: 100vh;
```
flodolo: Do we really want to use something that is not even supported in the current Firefox ESR?
http://caniuse.com/#feat=viewport-units

web/style/transvision.css
```diff
@@ -528,37 +524,113 @@ fieldset {
...
+    box-shadow: 0 0 1px 0 rgb(255, 255, 255), inset 0 1px 0 0 rgba(0, 0, 0, 0.03);
+    vertical-align: middle;
+}
+
+input[type="checkbox"]:not(:disabled) + label:hover::before {
+    border-color: #0095dd;
+}
+
+input[type="checkbox"]:checked + label:before {
+    background-image: url("/img/check.svg#check"), linear-gradient(#fff, rgba(255, 255, 255, 0.8));
```
flodolo: url without quotes (also in the rest of the file)

web/style/transvision.css
TheoChevalier: On your link, it is supported since Firefox 19

flodolo: D'oh, I assumed it stopped at 46, confusing table 🐻

web/style/transvision.css
TheoChevalier: you’re right about this one, but the url("data:image…") still requires it

flodolo:
I haven't fully looked at the code yet, but some notes while using it on my machine.

First thing: great job, It looks great.

The font looks really bad when small and bold, almost out of focus. I'm not a huge fan of it.
schermata 2016-07-14 alle 16 18 29

Main wrapper needs a bit more horizontal padding, the Clipboard icon almost overlaps with the scrollbar.

Let me know when you think it's ready for a deeper review.

TheoChevalier:
Ugh, that looked indeed bad… Good catch! Those combination of font + weight were computed by the browser, and while that looked good on retina screens that didn’t work on other screens. I think I’ve fixed them all, if you still see blurry fonts somewhere, let me know. I’ll look at the other issues later, thanks for the preliminary feedback :)

TheoChevalier:
Do you have a screenshot of the clipboard almost overlapping with the scrollbar please? I’d like to make sure we’re talking about the same thing with the same config

flodolo:
schermata 2016-07-15 alle 07 46 16

Basically there should be a little more space between the beginning/end of the table and the end of the window (the light blue stripe should be larger).

flodolo:
Titles still feel a bit foggy, also wonder if they should be bold

schermata 2016-07-15 alle 07 49 18

TheoChevalier:
@flodolo Lets consider this is ready for review.
release-the-wee-kraken

 Still haven’t fixed the titles, though. I think making them a little bolder would fix it since I see the same kind of artifacts with the fonts on prod. I’d just really wish we could make them a little thinner…

flodolo:
It looks great, and I can't really spot any issues with the code or testing.

My biggest concern is the font readability, considering how thin the font looks in some cases. Also need to check locales like Arabic, it looks super small.

The `Select All` button in TMX is unstyled: wanted or bug?

flodolo:
I wonder if we can improve this string

schermata 2016-07-27 alle 11 39 11

TheoChevalier:
Maybe we could move it to the placeholder like this?

capture d ecran 2016-07-27 a 12 27 37

flodolo:
Not really, we would lose it after the first search since the field is already filled.

TheoChevalier:
Yeah, good point. We could also play with the search button (tooltip or changing the value).

Regarding Arabic, Lizie & Mohamed confirmed this is as readable as prod, and that it’s the common font size for arabic websites.

flodolo:
Using the button's label seems hard (space constraints).

If the Select All button is expected, I think this can be merged and we can test further on dev servers.

TheoChevalier:
The Select All button is expected, but we could still improve it and create an alternative theme (I just felt it should not be a big blue button like the submit one)

flodolo:
I agree on the color, but it really looks out of place there. We could use the same style, just smaller and with a different background color (it could still be gray).

flodolo:
Button looks good. As for the text under the search input, setting it to 0.9 em looks already a lot better, but it would be enough as a fix.

Ready to merge when you want.

TheoChevalier:
Thanks, I felt we could also reduce the bottom margin. I’d say it could be merged if the last commit looks good

TheoChevalier:
comparison

Issue #741, opened 2016-05-01
This URL shows a blank page


TheoChevalier:
This should not show a blank page https://transvision.mozfr.org/string/?entity=toolkit/chrome/global/extensions.properties:csp.error.illegal-host-wildcard&repo=aurora

Works fine with central, shows "Can’t find this string" on beta as expected, but crashes on aurora.

flodolo:
It works just fine on my VM, which is now running Ubuntu 16.04 and php7 (PHP 7.0.4-7ubuntu2)
https://transvision.flod.org/string/?entity=toolkit/chrome/global/extensions.properties:csp.error.illegal-host-wildcard&repo=aurora

Does the log contain any info?

TheoChevalier:
Looks like we’ve got this corruption error once more:

TheoChevalier:
For the record, we got a new one today
1: Got error 'PHP message: PHP Parse error:  syntax error, unexpected 'verkko' (T_STRING), expecting ']' in /srv/mozfr01-data/transvision/TMX/fi/cache_fi_aurora.php on line 8589

flodolo:
OK, can you review this one and merge so it's out of the way?

flodolo:
Ah, didn't found this one when filed the new PR. This will be fixed by #776

TheoChevalier:
Sure, will merge today. I will keep monitor prod until we do a release

PR #776, opened 2016-07-28
Fixes #775 and #741 - Run PHP linter on cache files

### LINKED ISSUE #775: Verify TMX cache files (php linter)
In case the file doesn't pass `php -l`, we should rebuild the cache.

flodolo:
Also make sure to always build en-US Gaia cache (unrelated bug).

To test, it's enough to manually break some cache files.

It doesn't seem to have a visible negative impact on the running time.

### DIFF
--- a/app/scripts/glossaire.sh
+++ b/app/scripts/glossaire.sh
@@ -168,7 +168,7 @@ function updateStandardRepo() {
 
     updated_english=false
 
-    # Store md5 of the existing cache before updating the repositories
+    # Store md5 of the existing en-US cache before updating the repositories
     cache_file="${root}TMX/en-US/cache_en-US_${repo_name}.php"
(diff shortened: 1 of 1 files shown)


app/scripts/glossaire.sh
```diff
@@ -249,6 +257,14 @@ function updateStandardRepo() {
             then
                 echored "Cache doesn't exist for ${repo_name}/${locale_code}"
                 updated_locale=1
+            else
+                php -l $cache_file 2>&1 1>/dev/null
```
TheoChevalier: Is this supposed to echo the parse error in the console? If it’s expected, everything looks good

flodolo: Yes, it's expected to only output errors to the console

PR #780, opened 2016-07-29
Fixes #781 - Add force-tmx and no-update options to glossaire.sh

### LINKED ISSUE #781: Move forceTMX and updaterepo to command line in glossaire.sh

flodolo:


TheoChevalier:
That works great :) My only concern is that I totally see myself making typos with this (potentially) long command, or even typing `./glossaire.sh -help` because I’m mixing up syntax with this one https://github.com/mozfr/transvision/blob/master/start.sh#L19 .

Do you think we could add something like `Is this the command you would like to execute? [Y/n]` right after the command summary? Something generic like that would do it I suppose (and we already require bash 4.x) http://stackoverflow.com/a/3232082/1601864

TheoChevalier:
oh wait, that wouldn’t work with the cronjob… facepalm

flodolo:
I'd prefer to not add more complexity than necessary to this check, especially considering that this script is run only on "production" servers. Also put the echo at the beginning to make it easier to spot errors in the command.

flodolo:
Taking the `That works great` part and merging, we can evaluate to make further changes in case.

PR #783, opened 2016-07-29
Prevent the logo from slightly resizing on hover

TheoChevalier:
Small detail, but quite frustrating and I finally found why


flodolo:
Weird, I wonder if it's noticeable only on retina displays (can't see it from my computer). Anyhow, it looks like a safe change to take 🐼

PR #784, opened 2016-07-30
Improve a11y and make results easier to read

TheoChevalier:
This PR addresses the feedback received over the last couple days.

Fira Sans Regular is too bold, Fira Sans Light is hard to read for some people, let’s try with Fira Sans Book. We’re running out of alternative weight, so this better be the good one ;) I think it’s very similar to what we have on prod. I also hardcoded the font-size in pixels, 0.9em makes it ~12.96px, and 13px seem to looks better (tried on a non hi-DPI screen)


flodolo:
Results look definitely more readable to me, also no issue in using px.

I'll take a look at the ESLint failure later.

PR #785, opened 2016-07-31
Fixes #732: Add a menu to /channelcomparison

### LINKED ISSUE #732: Add a menu to /channelcomparison
Follow up of recent /channelcomparison revamp to include added strings, we would need a menu with link to the different sections (quite hard to scroll down to find them currently)

TheoChevalier:


flodolo:
Definitely + for the feature, but I'm not really in love with that border, I would go with simple bullets.

Also not completely sure about the word "views", but can't think of anything better.

TheoChevalier:
> Definitely + for the feature, but I'm not really in love with that border, I would go with simple bullets.

Yeah, that was… border/line.

bear-pun_o_854713

flodolo:
I'll act like I didn't read the last comment 🍷 and just merge this into oblivion 💫

PR #787, opened 2016-08-02
Fix highlighting

TheoChevalier:


app/classes/Transvision/Strings.php
```diff
@@ -103,6 +103,45 @@ public static function multipleStringReplace($replacements, $string)
...
+    }
+
+    /**
+     * Highlight special characters in the string
+     *
+     * @param  string $string Source text
+     * @return string Same string with specific sub-strings in <span>
+     *                       elements for styling with CSS
+     */
+    public static function highlightSpecial($string)
```
flodolo: I think it would be better to only have one function, and pass the "whitespace" as a parameter. If you want whitespaces, pass the parameter and that will add one line to `$replacements`. Too much code duplication right now..

flodolo:
Can you double check if this has any impact performance-wise on large views?

TheoChevalier:
I monitored perfs on comparison view, that did not seem to have an impact. Let me check with other views

TheoChevalier:
Tried with the main view and big query, that do not seem to impact perfs. However this is very fast on this machine, so there might be a small difference that I can’t spot here.

/?recherche=te&repo=aurora&sourcelocale=en-US&locale=fr&search_type=strings

Master:
[Tue Aug  2 16:35:47 2016] Memory peak: 25165824 (24MB)
[Tue Aug  2 16:35:47 2016] Elapsed time (s): 0.0862
[Tue Aug  2 16:35:49 2016] Memory peak: 25165824 (24MB)
[Tue Aug  2 16:35:49 2016] Elapsed time (s): 0.0903
[Tue Aug  2 16:35:51 2016] Memory peak: 25165824 (24MB)
[Tue Aug  2 16:35:51 2016] Elapsed time (s): 0.0814

This branch:
[Tue Aug  2 16:34:51 2016] Memory peak: 25165824 (24MB)
[Tue Aug  2 16:34:51 2016] Elapsed time (s): 0.081
[Tue Aug  2 16:34:53 2016] Memory peak: 25165824 (24MB)
[Tue Aug  2 16:34:53 2016] Elapsed time (s): 0.0802
[Tue Aug  2 16:34:54 2016] Memory peak: 25165824 (24MB)
[Tue Aug  2 16:34:54 2016] Elapsed time (s): 0.0812

app/classes/Transvision/Strings.php
```diff
@@ -103,6 +103,32 @@ public static function multipleStringReplace($replacements, $string)
     }
 
     /**
+     * Highlight special characters in the string
+     *
+     * @param  string $string Source text
+     * @param  string $white_space Optional param to specify if we need to
```
flodolo: Should we use a boolean? It seems more logical (not sure if my first comment through you off here).

Not sure if you want to use $exclude_whitespaces (default True), to exclude changing too many existing calls.

TheoChevalier: haha, yeah, I used a boolean initially then thought you were suggesting to use a string. Thanks, will update that. Yeah, $exclude_whitespaces should be better, since there’s a single call highlight them

flodolo:
Everything looks good, thanks.

PR #788, opened 2016-08-02
Set font fallback on Arial for unsupported locales, prevent the fallb…

TheoChevalier:
…ack font to be bold

'Arial' was here for a reason :)

### DIFF
--- a/web/style/transvision.css
+++ b/web/style/transvision.css
@@ -49,6 +49,15 @@
     font-style: normal;
 }
 
+@font-face {
+    font-family: 'Fira Sans';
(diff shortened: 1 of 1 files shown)


flodolo:
I confess that I'm starting to be a bit concerned about how many fonts we're loading (we're approaching 1 MB), and if we should go back to use a smaller set of fonts.

TheoChevalier:
> I confess that I'm starting to be a bit concerned about how many fonts we're loading (we're approaching 1 MB), and if we should go back to use a smaller set of fonts.

I’m not adding a new one, just reusing it with a different name

flodolo:
Completely missed that.

flodolo:
Tested and everything looks good, thanks.

PR #792, opened 2016-08-04
Transliterate button

TheoChevalier:
Adding changes on top of #754 to add the button on entity view, so that we ship it in today’s release :)  Will comment when it’s ready for review

### DIFF
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -157,8 +157,9 @@ class ShowResults
         $request = new API(parse_url(API_ROOT . "transliterate/$locale/$string"));
         $json = include MODELS . 'api/transliterate.php';
 
-        return $json[0];
+        return isset($json[0]) ? $json[0] : 'Function not available';
     }
+
     /**
      * Return search results in a repository on strings/entities for the API
      *
@@ -316,7 +317,7 @@ class ShowResults
 
             /*
                 Find if we need to transliterate the string.
-                The string gets transliterated if the target local is serbian,
+                The string gets transliterated if the target local is Serbian,
                 if we aren't in the 3locales view and if we have a $target_string
             */
             $transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string && $target_string != '@@missing@@';
@@ -349,6 +350,7 @@ class ShowResults
             if ($transliterate) {
                 $transliterated_string = htmlspecialchars($transliterated_string);
                 $transliterated_string = Utils::highlightString($transliterated_string);
+                $transliterated_string = Strings::highlightSpecial($transliterated_string);
             }
 
             if ($extra_locale) {
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -145,6 +145,21 @@ class ShowResults
     }
 
     /**
+     * Use the API to return a transliterated string
+     *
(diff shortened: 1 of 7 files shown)


TheoChevalier:
Removed secureText, because we’re already using it here https://github.com/mozfr/transvision/pull/792/files#diff-9072e93e713a4782f5407ac9e3dbc369R34

However, we don’t seem to be using it in ShowResults, I assume we should?

TheoChevalier:
Ready for review, I’ll manually squash then merge without auto-squashing, so that we keep two commits

app/classes/Transvision/ShowResults.php
```diff
@@ -296,22 +312,47 @@ public static function resultsTable($search_object, $search_results, $page)
                 $target_string2 = '';
             }
 
+            $string_id = md5($key . mt_rand());
+            $regular_string_id = 'string_' . $string_id;
+
+            /*
+                Find if we need to transliterate the string.
+                The string gets transliterated if the target local is serbian,
```
flodolo: nit: Serbian

app/classes/Transvision/ShowResults.php
```diff
@@ -145,6 +145,22 @@ public static function getTranslationMemoryResults($strings, $search, $max_resul
...
+     * @param string $locale The target locale
+     *
+     * @return string The string transliterated into the target locale
+     */
+    public static function getTransliteratedString($string, $locale)
+    {
+        $request = new API(parse_url(API_ROOT . "transliterate/$locale/$string"));
+        $json = include MODELS . 'api/transliterate.php';
+
+        return $json[0];
```
flodolo: I still think this should check if $json actually has an element, and figure out what to do if the string returned is empty.

app/inc/constants.php
```diff
@@ -17,6 +17,7 @@
 define('CACHE_ENABLED', isset($_GET['nocache']) ? false : true);
 define('CACHE_PATH',    INSTALL_ROOT . 'cache/');
 define('APP_SCHEME',    isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off' ? 'https://' : 'http://');
+define('API_ROOT',  APP_SCHEME . htmlspecialchars($_SERVER['HTTP_HOST'], ENT_QUOTES, 'UTF-8') . '/api/v1/');
```
flodolo: Follow existing alignment?

flodolo:
Can't find anything wrong during testing. I feel like we should be more cautious with errors, especially considering the transliterate function requires a component that is not part of default PHP installations.

TheoChevalier:
If last commit looks good, I’m ready to squash & merge

app/classes/Transvision/ShowResults.php
```diff
@@ -145,6 +145,22 @@ public static function getTranslationMemoryResults($strings, $search, $max_resul
...
+     * @param string $locale The target locale
+     *
+     * @return string The string transliterated into the target locale
+     */
+    public static function getTransliteratedString($string, $locale)
+    {
+        $request = new API(parse_url(API_ROOT . "transliterate/$locale/$string"));
+        $json = include MODELS . 'api/transliterate.php';
+
+        return isset($json[0]) ? $json[0] : '@@missing@@';
```
flodolo: Call to this method is after checks on `@@missing@@`.  Let's either use an empty string, or something like 'Function not available'.

PR #793, opened 2016-08-04
Fix missing background on horizontal scroll

TheoChevalier:
This fixes the missing background here https://transvision-beta.mozfr.org/3locales/?recherche=test&repo=aurora&sourcelocale=en-US&locale=fr&locale2=fr&search_type=strings

I tried to style the table differently to get rid of the overflow, had good results with  table-layout: fixed; and break-words, but not good enough and that seem a bad idea to touch that today.

### DIFF
--- a/web/style/transvision.css
+++ b/web/style/transvision.css
@@ -62,6 +62,7 @@
 
 html {
     min-height: 100%;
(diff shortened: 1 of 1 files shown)


flodolo:
It works 🐷

PR #801, opened 2016-08-06
Fixes #800 - Channel comparison is showing wrong last changed string, labels overlapping

### LINKED ISSUE #800: Channel Comparison shows wrong original string
See https://transvision.mozfr.org/channelcomparison/?locale=cs&chan1=central&chan2=beta - second row for (toolkit • chrome • global • languageNames.properties or). There should be "Udijština" instead of "Vydavatel certifikátu partnera nebyl rozpoznán.", which is duplicated from the first row (I guess some for cycle issue).

flodolo:


flodolo:
I've reverted the change to unlocalized_words, it gives a completely different set of results without the variable passed as reference, and that's something that I don't plan to look at :-)

TheoChevalier:
I don’t know why I was convinced it had to be passed as reference… My bad.

And yes, changing things in unlocalized_words is very likely to alter either results or performance :/

TheoChevalier:
That looks good, merging

PR #803, opened 2016-08-07
Fixes #802 - @@missing@@ shown instead of "Missing string" for string…

### LINKED ISSUE #802: @@missing@@ shown instead of "Missing string" for string searches
https://transvision.mozfr.org/?recherche=use+in+console&repo=aurora&sourcelocale=en-US&locale=de&search_type=strings has two results with "@@missing@@" as German translation. Transvision even run checks against it ("No final dot? Small string?").

TheoChevalier:
… searches

Not sure if we want to do the change for entity view. The same could happen if we decide to highlight @ characters for instance.


flodolo:
Wow, this is really bad 😭 

As suggested in the past, I think we should start releasing more often, updating the changelog only on major releases (this one would be a 5.0.1). Just tag the release and pull it on the server.

Returning to the subject: why don't we fix it upstream? It feels a lot safer, and would fix all the view.
1) `Utils::markString` and `Utils::highlightString` should probably be in Strings, not Utils

2) We can add a check at the beginning of `markString`
```PHP
        if ($haystack == '@@missing@@') {
            return $haystack;
        }
```

3) If we want to highlight @, we can add a similar check in `highlightString`

TheoChevalier:
I was considering a 5.0.1 too, doing it without a changelog works for me.

1) And yeah, fixing it upstream sound better.

2) Sounds good

3) I don’t think we want, it’s just we might be doing things like that in the future, who knows

app/classes/Transvision/ShowResults.php
```diff
@@ -320,42 +323,42 @@ public static function resultsTable($search_object, $search_results, $page)
                 The string gets transliterated if the target local is Serbian,
                 if we aren't in the 3locales view and if we have a $target_string
             */
-            $transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string && $target_string != '@@missing@@';
+            $transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string && ! $target_string_missing;
```
flodolo: I would leave the original check at this point.

PR #804, opened 2016-08-07
Fixes #739 - Add php-mbstring to dependencies

### LINKED ISSUE #739: Add php-mbstring to dependencies
After upgrading Ubuntu to 16.04, Transvision fails launching because php-mbstring package is missing (now opt-in with php7).
Without it, app crashes here https://github.com/mozfr/transvision/blob/master/app/inc/init.php#L8

I don’t see any other change, so far

SkySymbol:


TheoChevalier:
Issue #739 was only about adding a line to Readme.md, I think we should keep using the PHP provided package

PR #806, opened 2016-08-19
Fix warnings in mainsearch.php and search_counter.php

flodolo:


flodolo:
The alternative to the last commit is to use `align_equals`, but it gives awful results in some parts of the codebase. Also, personally not a fan of aligning equal signs in assignments.

TheoChevalier:
LGTM, sorry for the delay! 
I think I liked the extra spaces in some cases, but I’m fine with the current. I’m tempted not to squash your commits, given the broad php-cs-fixer changes, proceed as you wish ;)

PR #814, opened 2016-09-26
Create QA view for Empty strings (fixes #813)

### LINKED ISSUE #813: Create QA view for empty strings
I noticed several locales doing strange things with empty strings, mostly to make tools happy.

I plan to create a view that shows empty strings in English and locale, now that we can differentiate them.

flodolo:


TheoChevalier:
Getting these on central, but will check if it’s because locales are not up-to-date:

[Mon Sep 26 14:52:22 2016] PHP Notice:  Undefined index: toolkit/chrome/mozapps/help/help.dtd:fullZoomEnlargeCmd.commandkey3 in /Users/tchevalier/moz/transvision/app/models/empty_strings.php on line 47
[Mon Sep 26 14:52:22 2016] PHP Notice:  Undefined index: toolkit/chrome/mozapps/help/help.dtd:fullZoomReduceCmd.commandkey2 in /Users/tchevalier/moz/transvision/app/models/empty_strings.php on line 47
[Mon Sep 26 14:52:22 2016] PHP Notice:  Undefined index: toolkit/chrome/mozapps/help/help.dtd:fullZoomResetCmd.commandkey2 in /Users/tchevalier/moz/transvision/app/models/empty_strings.php on line 47

flodolo:
Interesting. I didn't put a check on that line because we don't store obsolete strings, so a locale shouldn't be expected to have extra keys. I guess I can add the check like the previous cycle, just in case.

TheoChevalier:
Looks like we need to escape html tags, they are interpreted at the bottom of http://localhost:8082/empty-strings/?locale=de&repo=central

TheoChevalier:
(And the errors above were definitely due to l10n repos not being up-to-date)

app/models/empty_strings.php
```diff
@@ -0,0 +1,65 @@
...
+            'reference'   => $reference_strings[$string_id],
+            'translation' => $value,
+        ];
+    }
+}
+
+unset($reference_strings);
+unset($locale_strings);
+
+if (count($empty_strings > 0)) {
```
TheoChevalier: Hm, this does not seem to work, I get both the filter block and "No strings found" with /empty-strings/?locale=eo&repo=central

TheoChevalier:
Maybe we should separate results into two groups: when the source is empty, and when the locale is empty? Or make the columns sortable?

flodolo:
My first approach was to have 2 separate tables, but then I decided to only show the strings that diverge (i.e. not both empty), and given the small number of results it seemed OK to only have one table.

Added the missing .js to make them sortable

app/models/empty_strings.php
flodolo: Yeah, if I only I were able to put the parenthesis where they belong… (fixing)

TheoChevalier:
Thanks, looks good 👍

PR #820, opened 2016-10-13
Use external library for p12n-extract

### LINKED ISSUE #819: Use p12n-extract script from external source
Productization is changing too fast for Transvision these days
https://transvision.mozfr.org/productization/?locale=it&product=browser

We should clone p12n-extract git repository instead of having a static version of the script

flodolo:
Fixes #819


flodolo:
@TheoChevalier 
Sanity check?

If you want to test it: if you check Italian for Firefox and Thunderbird, it shouldn't have searchplugins on Nightly, while it should have them after running glossaire with this patch.

flodolo:
Thanks, I'll leave it run on my VM for a few days and see if anything strange happens.

PR #825, opened 2016-11-09
Stop changing whitespaces in searched strings

### LINKED ISSUE #824: Perfect match search sometimes fails when the string is available
For example
https://transvision.mozfr.org/?recherche=%3Cp%3E%26brandShortName%3B+can%E2%80%99t+load+this+page+for+some+reason.%3C%2Fp%3E&repo=aurora&sourcelocale=en-US&locale=eu&search_type=strings
vs.
https://transvision.mozfr.org/?recherche=%3Cp%3E%26brandShortName%3B+can%E2%80%99t+load+this+page+for+some+reason.%3C%2Fp%3E+&repo=aurora&sourcelocale=en-US&locale=eu&search_type=strings&perfect_match=perfect_match

https://transvision.mozfr.org/?recherche=section.&repo=aurora&sourcelocale=en-US&locale=eu&search_type=strings
vs.
https://transvision.mozfr.org/?recherche=section.&repo=aurora&sourcelocale=en-US&locale=eu&search_type=strings&perfect_match=perfect_match

flodolo:
The idea is that we change the searched string a log, by dropping trailing/leading whitespaces, and collapsing multiple spaces. I don't think it makes a lot of sense.

You can get an idea by looking at this page and searching some of HTML strings
https://transvision.mozfr.org/consistency/?locale=eu&repo=aurora&filter=all

Note: the string appear on multiple lines because the `highlight` class is acting as `<pre>` now, so it's expected.

I've updated the mtrim function, but it's currently unused. Not sure if we want to get rid of it completely.

I've been testing this for a bit, and couldn't find any evident issue. This also fixes #824

### DIFF
--- a/app/classes/Transvision/Search.php
+++ b/app/classes/Transvision/Search.php
@@ -132,7 +132,7 @@ class Search
      */
     public function setSearchTerms($string)
     {
-        $this->search_terms = trim($string);
+        $this->search_terms = $string;
         $this->regex_search_terms = $this->search_terms;
         $this->updateRegex();
 
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -173,7 +173,7 @@ class ShowResults
         $output = [];
 
         $clean_string = function ($string) {
-            return trim(htmlspecialchars_decode($string, ENT_QUOTES));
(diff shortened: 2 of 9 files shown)


flodolo:
Merging per IRC discussion 👍

PR #826, opened 2016-11-14
Fixes #823 - Add a link to Pootle or Pontoon on strings

### LINKED ISSUE #823: Add a link to Pootle or Pontoon on strings
It would be great to have a link to open the editor (Pootle, Pontoon) directly, like we link to sources.

SkySymbol:
I've got the list of locales for `isPresentInPootle` and `isPresentInPontoon` from those links:

- https://mozilla.locamotion.org/projects/firefox/
- https://pontoon.mozilla.org/projects/firefox-aurora/


app/classes/Transvision/ShowResults.php
```diff
@@ -322,6 +323,12 @@ public static function resultsTable($search_object, $search_results, $page)
             */
             $transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string && $target_string != '@@missing@@';
 
+            if ($current_repo === 'aurora' && ShowResults::isPresentInPootle($locale2)) {
+                $link = "https://mozilla.locamotion.org/$locale2/firefox/translate/$entityPath.po?search=$strings[1]&sfields=source&sfields=target";
```
flodolo: Let's use a better variable name here, e.g. edit_link
Also,`"{$variable_name"` in strings

app/classes/Transvision/ShowResults.php
```diff
@@ -479,7 +486,14 @@ public static function resultsTable($search_object, $search_results, $page)
                     <div dir='ltr' class='result_meta_link'>
                       <a class='source_link' href='{$locale2_path}'>
                         &lt;source&gt;
-                      </a>
+                      </a>";
+            if ($link) {
+                $table .= "
+                            <a class='bug_link' target='_blank' href='$link'>
```
flodolo: Create a new class name and update the CSS

app/classes/Transvision/ShowResults.php
flodolo: * Let's use a better variable name here, e.g. edit_link.
* `"{$variable_name"` in strings.
* Search for the string ID, not the string.

app/classes/Transvision/ShowResults.php
```diff
@@ -567,4 +581,286 @@ public static function buildErrorString($source_string, $target_string)
...
+     *
+     * @param string $locale String the locale we want to check
+     *
+     * @return bool true if the locale is present is Pootle and has a project
+     *              'firefox', false otherwise.
+     */
+    public static function isPresentInPootle($locale)
+    {
+        static $localesInPootle = [
+            "ach",
```
flodolo: There's no need to define the variable as static inside a function, also use single quotes when double quotes are not needed.

flodolo:
Let's try to improve the view part as much as possible. I'll probably open a twin PR to get the list of locales from the external sources.

SkySymbol:
Ok. I'll fix the points you mentioned. Let's put this PR in standby until we find a better way to get those lists.

flodolo:
You can use `Project::getLocaleTool` now. Make sure to only query it one time at the beginning of the view.

Also, run `setup.sh` once to create the local configuration files.

SkySymbol:
Thank you

app/classes/Transvision/ShowResults.php
```diff
@@ -479,7 +488,14 @@ public static function resultsTable($search_object, $search_results, $page)
                     <div dir='ltr' class='result_meta_link'>
                       <a class='source_link' href='{$locale2_path}'>
                         &lt;source&gt;
-                      </a>
+                      </a>";
+            if ($edit_link) {
```
flodolo: You need to set $edit_link to a null string, otherwise there will be a ton of warnings for locales not working in tools.

app/classes/Transvision/ShowResults.php
```diff
@@ -321,6 +325,11 @@ public static function resultsTable($search_object, $search_results, $page)
                 if we aren't in the 3locales view and if we have a $target_string
             */
             $transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string && $target_string != '@@missing@@';
+            if ($current_repo === 'aurora' && $toolUsedByTargetLocale == 'locamotion') {
+                $edit_link = "https://mozilla.locamotion.org/{$locale2}/firefox/translate/{$entityPath}.po?search={$strings[1]}&sfields=source&sfields=target";
```
flodolo: As suggested in the original issue, let's search for the entity and `sfields=locations`.
You can store the exploded $key and refer to [0] and [1] instead of creating two variables.

web/style/transvision.css
```diff
@@ -451,6 +451,10 @@ td .result_meta_link .bug_link {
     color: #404d6c;
 }
 
+td .result_meta_link .edit_link {
+    color: #404d6c;
```
flodolo: I would use a different color, also needs margin on the left.

app/classes/Transvision/ShowResults.php
```diff
@@ -479,7 +488,14 @@ public static function resultsTable($search_object, $search_results, $page)
                     <div dir='ltr' class='result_meta_link'>
                       <a class='source_link' href='{$locale2_path}'>
                         &lt;source&gt;
-                      </a>
+                      </a>";
+            if ($edit_link) {
+                $table .= "
+                            <a class='edit_link' target='_blank' href='{$edit_link}'>
```
flodolo: Add a tooltip, like "Open this string in X", with X name of the tool.

app/classes/Transvision/ShowResults.php
```diff
@@ -321,6 +325,11 @@ public static function resultsTable($search_object, $search_results, $page)
                 if we aren't in the 3locales view and if we have a $target_string
             */
             $transliterate = $locale2 == 'sr' && ! $extra_locale && $target_string && $target_string != '@@missing@@';
+            if ($current_repo === 'aurora' && $toolUsedByTargetLocale == 'locamotion') {
```
flodolo: `== 'aurora'` is more than enough (same below)

app/classes/Transvision/ShowResults.php
```diff
@@ -321,6 +325,11 @@ public static function resultsTable($search_object, $search_results, $page)
...
+                $edit_link = "https://mozilla.locamotion.org/{$locale2}/firefox/translate/{$fileAndRawString[0]}.po?search={$fileAndRawString[1]}&sfields=locations";
```
flodolo: This link is not working, you need a `#` before `search`.

Example: https://mozilla.locamotion.org/eu/firefox/translate/browser/chrome/browser/aboutPrivateBrowsing.dtd.po#search=aboutPrivateBrowsing.info.cookies&sfields=locations

app/classes/Transvision/ShowResults.php
```diff
@@ -479,7 +488,15 @@ public static function resultsTable($search_object, $search_results, $page)
                     <div dir='ltr' class='result_meta_link'>
                       <a class='source_link' href='{$locale2_path}'>
                         &lt;source&gt;
-                      </a>
+                      </a>";
+            if (isset($edit_link)) {
+                $table .= "
+                            &nbsp;
+                            <a class='edit_link' target='_blank' href='{$edit_link}'>
+                            &lt;edit&gt;
```
flodolo: If we don't want to put a tooltip, given it's bad for accessibility, I would make this link "<edit in Pootle>" and "<edit in Pontoon>", depending on the tool. You can set the text in the if cycle where you create the link.

SkySymbol:
Thank you for the review :-)

flodolo:
Looks great, thanks.

We'll be missing the link in the 3 locales view, but I won't to open a discussion about that.

PR #829, opened 2016-11-19
Use compare-locales for parsing product repositories

flodolo:
Expected differences in output:
* Silme keeps trailing whitespaces in .properties, compare-locale does not. The new behavior is correct.
* compare-locales reads keys like `mobile/android/chrome/aboutLogins.dtd:brandDTD`. That's correct.
* Previous script completely ignored /b2g files.

Corresponding upstream PR, with tests & C.
https://github.com/flodolo/tmx_maker/pull/6

### DIFF
--- a/app/scripts/setup.sh
+++ b/app/scripts/setup.sh
@@ -74,26 +74,32 @@ function createSymlinks() {
     esac
 }
 
-function checkoutSilme() {
-    # Check out SILME library to a specific version (0.8.0)
-    if [ ! -d $libraries/silme/.hg ]
+function setupExternalLibraries() {
+    # Check out or update compare-locales library
+    version="RELEASE_1_2_1"
+    if [ ! -d $libraries/compare-locales/.hg ]
     then
-        echogreen "Checking out the SILME library into $libraries"
+        echogreen "Checking out compare-locales in $libraries"
         cd $libraries
-        hg clone https://hg.mozilla.org/l10n/silme -u silme-0.8.0
+        hg clone https://hg.mozilla.org/l10n/compare-locales -u $version
+        cd $install
+    else
+        echogreen "Updating compare-locales in $libraries"
+        cd $libraries/compare-locales
+        hg pull -r default --update
+        hg update $version
         cd $install
     fi
-}
 
-function setupP12nExtract() {
+    # Check out or update external p12n-extract library
     if [ ! -d $libraries/p12n/.git ]
     then
-        echogreen "Checking out the p12n-extract library into $libraries"
+        echogreen "Checking out the p12n-extract library in $libraries"
         cd $libraries
         git clone https://github.com/flodolo/p12n-extract/ p12n
         cd $install
     else
-        echogreen "Updating the p12n-extract library into $libraries"
+        echogreen "Updating the p12n-extract library in $libraries"
         cd $libraries/p12n
         git pull
         cd $install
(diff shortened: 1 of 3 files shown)


flodolo:
One more note: we could live without storing a JSON version of the cache (adding `--output php` to the call in glossaire.sh), but that means that we wouldn't be removing obsolete strings (i.e. strings not available in the reference language).

In a l20n world, we'll need to remove that check anyway, since locales can have more strings than reference.
https://github.com/flodolo/transvision/blob/21edc75eaad0891e16fc90d6365c075a9937e479/app/scripts/tmx_products.py#L150-L161

This version of tmx_products.py is already designed to work in the cross-channel scenario, not yet ready for l20n (no parser).

app/scripts/setup.sh
```diff
@@ -74,18 +74,24 @@ function createSymlinks() {
     esac
 }
 
-function checkoutSilme() {
-    # Check out SILME library to a specific version (0.8.0)
-    if [ ! -d $libraries/silme/.hg ]
+function setupExternalLibraries() {
+    # Check out or update compare-locales library
+    version="RELEASE_1_2_1"
```
TheoChevalier: silme was never updated and we needed a specific version, but maybe we will want to update compare-locales without having to do a new release? Of course we don’t want to automatically use the latest without testing it before, but maybe we should move the version number into the config file? We could then use it in both setup.sh and tmx_product.py

app/scripts/setup.sh
```diff
@@ -74,18 +74,24 @@ function createSymlinks() {
...
+    if [ ! -d $libraries/compare-locales/.hg ]
     then
-        echogreen "Checking out the SILME library into $libraries"
+        echogreen "Checking out the compare-locales into $libraries"
```
TheoChevalier: “Checking out compare-locales into $libraries”?

app/scripts/setup.sh
flodolo: I'm not sure storing it in config would be a good solution, mostly because we wouldn't have clear control over it (obsolete config -> obsolete library version, also potentially non existent version).

I don't expect compare-locales to be updated frequently, I just want to make sure we have the version we want (e.g. 1.2 had bugs, I want to make sure we have 1.2.1). The cost in terms of time is acceptable, since it's run only in setup.sh

If compare-locales changes significantly, we would still need an update to the main Python script.

app/scripts/setup.sh
flodolo: Meh, fixing 💩

app/scripts/setup.sh
flodolo: One more note: I though about that, but I couldn't come up with an acceptable solution.

We could checkout tmx_products as an external library, like we do with p12n-extract, but that wouldn't really solve the issue with compare-locales (I can't run hg pull every time the library it's opened, only clone the repo if it's missing).

The alternative would be to install compare-locales as a proper Python package (changing the path is frowned upon), and maybe use a virtual env. It would be much cleaner, but it would also introduce more dependencies on the server.

TheoChevalier: Yeah you’re right, we don’t want to debug instances where people would see “weird bugs” because they are using the wrong compare-locales version. I was only thinking about the prod server.
Let’s keep what you did, we can still revisit later if it happens we need to update it more frequently

flodolo: OK. Let me know when you've been able to give it a try, then we'll merge it and let it bake a bit on my dev server.

PR #833, opened 2016-11-27
Rework advanced search options, and default search behavior

### LINKED ISSUE #831: Change the way we search
I confess that I find Transvision search really confusing when it comes to non basic searches.

For example: I want to find all strings containing "set a cookie". I go ahead and search for it, and find something like 29k results! Why? Because we split the sentence in words, and then proceed to search for each unique word in it. And there's "a" in that sentence.

Perfect match doesn't fit, "Whole word" works in a pretty random way and fails at highlighting.

My suggestion would be to:
* Drop the "Whole word" checkbox, and replace it with "Distinct words", or "Search each word": if selected we should replicate the current behavior, searching for each word.
* Default would be to search for the entire sentence, not single words.

I have a WIP branch created just out of curiosity, but wanted to discuss about it before doing any more work and thinking on it.

It's also not a lot of code to change.

flodolo:
* Stop splitting words in a sentence by default when searching
* Add "Each Word" option to search all word in a sentence (old default behavior)
* Rename "Perfect march" to "Entire string", "Whole Words" to "Entire Words"
* Reword and reverse API promotion links
* Ignore words shorter than 2 characters when searching for each word
* Added tooltips to search options

Fixes #831

### DIFF
--- a/app/classes/Transvision/API.php
+++ b/app/classes/Transvision/API.php
@@ -15,7 +15,7 @@ use Monolog\Logger;
  * Calls are like this:
  * api/<version>/<service>/<repository>/<search type>/<source locale>/<target locale>/<url escaped search>/?optional_parameter1=foo&optional_parameter2=bar
  * Example for an entity search containing bookmark:
- * https://transvision.mozfr.org/api/v1/tm/release/entity/en-US/fr/bookmark/?case_sensitive=1
+ * https://transvision.mozfr.org/api/v1/tm/release/entity/en-US/fr/bookmark/?case_sensitive=case_sensitive
  * (tm = translation memory service)
  *
  * Example for the list of locales supported for a repo:
@@ -245,7 +245,7 @@ class API
 
                 break;
             case 'search':
-            // ex: /api/v1/search/string/central/en-US/fr/Bookmark/?case_sensitive=1
+            // ex: /api/v1/search/string/central/en-US/fr/Bookmark/?case_sensitive=case_sensitive
                 if (! $this->verifyEnoughParameters(7)) {
                     return false;
                 }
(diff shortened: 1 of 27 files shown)


flodolo:
I think this should be in a decent state (definitely more changes than I initially expected).

The only place where I left the search behavior unchanged is in suggestions, where I think it makes sense to search for each word by default.

flodolo:
For sake of testing, my dev server is currently running a dev branch with both #833 and #825 in it.

flodolo:
Moved tooltips: in some cases tooltips for options on the right would be cut off the screen.

TheoChevalier:
Sorry for the extra delay here, the bright side is that I’ve had plenty of time to test it locally ;)

I don’t think we should be able to keep Entire words checked when we check Entire string:
capture d ecran 2017-01-11 a 15 31 44
There’s no incidence on the query, but I think it can unnecessarily confuse people, especially in this part of the UI that is already confusing for many.
In addition to disabling Entire words in main.js, we should uncheck it.

capture d ecran 2017-01-11 a 15 32 03
Cut-off tooltips, but not sure it’s worth fixing.

Other than that, I’ve been using it over the last few weeks, it looks good to me 👍

flodolo:
Thanks. I'll try to take a look and fix them when I'm back.

I also couldn't help noticing the overlapping MozFR in the footer (last screenshot).

TheoChevalier:
Yeah, noticed it too…

Btw, deliberate choice to have two different styling for custom tooltips? I mean compared to the current one we have for the copy button

flodolo:
> Btw, deliberate choice to have two different styling for custom tooltips? I mean compared to the current one we have for the copy button

I kind of like to have an aspect similar to buttons for the ones in search, any strong opinion about using the same style for the clipboard notification?

PR #836, opened 2017-01-15
Make entity name selection easier - fixes #830

### LINKED ISSUE #830: Copy entity value or file name
Currently there is no easy way to copy the entity value or the file name from the table with the search results. When one tries to select the selection spans far from the start of the table cell. The problem is because `td .link_to_entity` is set to `float: left` for proper alignment. The simplest solution I can think of is to use `display: inline-block` so that the content (file name and entity) can be individually selected with the mouse.

As a workaround I use Caret Browsing.

TheoChevalier:


flodolo:
This looks good. On a separate issue, we should probably make the entity name a link in the accesskeys view.

PR #838, opened 2017-01-16
Replace @@missing@@ when source is missing

TheoChevalier:
Looks like there is another case we were not covering: https://transvision.mozfr.org/?recherche=Firefox+OS&repo=mozilla_org&sourcelocale=en-GB&locale=fr&search_type=strings&whole_word=whole_word#mozilla_org_firefox_os_tv.lang_9bf38c70

I wasn’t able to find a query where source is missing in entities results page, but made the change in results_entities.php anyway, just to be sure we’re covering the case

### DIFF
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -394,7 +394,9 @@ class ShowResults
             $meta_source = $meta_target = $meta_target2 = '';
 
             // If there is no source_string, display an error, otherwise display the string + meta links
-            if (! $source_string) {
+            if ($source_string == '@@missing@@') {
+                $source_string = '<em class="error">Warning: Source string is missing</em>';
+            } elseif (! $source_string) {
                 $source_string = '<em class="error">Warning: Source string is empty</em>';
             } else {
                 $meta_source = "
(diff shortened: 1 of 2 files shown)


flodolo:
I haven't looked at code yet, but technically source code can't be missing because of this
https://github.com/flodolo/tmx_maker/blob/master/tmx_products/tmx_products.py#L150-L161

But things will change with l20n, so worth fixing it anyway. I will review the PRs later this week, as soon as I'm back to a computer with a local install of Transvision.

TheoChevalier:
yeah, that’s true for products, but mozilla.org uses tmx_mozillaorg

flodolo:
mozilla.org's localization files can't have strings that are not available in en-US either, there's always a reference file to propagate changes from ;-)

Same for XLIFF, unless we start doing something funky with experiments.

TheoChevalier:
hm, true. "source" here is en-GB, not en-US, so that’s why we’re failing

flodolo:
You're absolutely right, I didn't check the link (blame jet-lagged brain 🍷 ).

On a side note, we should probably take a look at the extraction script and see if it can be improved.

PR #842, opened 2017-01-23
Rethink tooltips for search options

flodolo:
A few thoughts on the current implementation (tooltip on hover):
* It's annoying if you already know what the option does.
* It's broken on touch devices or, at least, on iOS. Spent an entire afternoon trying to figure out how to deal with it, but the series of events fired is completely inconsistent.

One possible alternative is, instead of showing a tooltip, opening a help page somewhere with more general instructions on how to use Transvision.

### DIFF
--- a/app/views/search_form.php
+++ b/app/views/search_form.php
@@ -129,7 +129,7 @@ $cookie_option = function ($cookie, $locale) use ($search) {
                     <label>Advanced Search options</label>
                     <a href="#" id="tooltip_search"><span>i</span></a>
                     <div class="tooltip" id="tooltip_search_text">
-                        <a href="#" id="tooltip_search_close" data-tooltip-id="tooltip_search_text" title="Close this panel">x</a>
+                        <a href="#" id="tooltip_search_close" data-tooltip-id="tooltip_search_text" title="Close this panel">×</a>
                         <h2>Case Sensitive</h2>
                         <p>If selected, search with the <strong>exact case</strong> used in the search query. By default, case is ignored.</p>
                         <h2>Each Word</h2>
@@ -157,7 +157,7 @@ $cookie_option = function ($cookie, $locale) use ($search) {
                                class="search_options"
                                <?=Utils::checkboxState($check['each_word'])?>
                         />
-                        <label>Each Word</label>
+                        <label for="each_word">Each Word</label>
                     </span>
                     <span>
                         <input type="checkbox"
@@ -167,7 +167,7 @@ $cookie_option = function ($cookie, $locale) use ($search) {
                                class="search_options"
                                <?=Utils::checkboxState($check['entire_string'])?>
                         />
-                        <label>Entire String</label>
+                        <label for="entire_string">Entire String</label>
                     </span>
                     <span>
                         <input type="checkbox"
@@ -177,7 +177,7 @@ $cookie_option = function ($cookie, $locale) use ($search) {
                                class="search_options"
                                <?=Utils::checkboxState($check['entire_words'])?>
                         />
-                        <label>Entire Words</label>
+                        <label for="entire_words">Entire Words</label>
                     </span>
                     <?php if ($check['t2t'] == 't2t') :?>
                     <input type="hidden"
--- a/app/views/search_form.php
+++ b/app/views/search_form.php
@@ -127,6 +127,18 @@ $cookie_option = function ($cookie, $locale) use ($search) {
 
                 <fieldset id="advanced_search">
                     <label>Advanced Search options</label>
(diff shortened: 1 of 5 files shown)


app/views/search_form.php
```diff
@@ -146,8 +157,7 @@ class="search_options"
                                class="search_options"
                                <?=Utils::checkboxState($check['each_word'])?>
                         />
-                        <label for="each_word" aria-describedby="tooltip_eachword">Each Word</label>
-                        <div id="tooltip_eachword" class="search_tooltip" role="tooltip">If selected, search for <strong>each word</strong> in the search query (at least 2 characters long). By default, the search query is used as a whole.</div>
+                        <label>Each Word</label>
```
TheoChevalier: Should be `<label for="each_word">Each Word</label>` (can’t be checked otherwise)

app/views/search_form.php
```diff
@@ -157,8 +167,7 @@ class="search_options"
                                class="search_options"
                                <?=Utils::checkboxState($check['entire_string'])?>
                         />
-                        <label for="entire_string" aria-describedby="tooltip_entirestring">Entire String</label>
-                        <div id="tooltip_entirestring" class="search_tooltip" role="tooltip">If selected, the <strong>entire string</strong> needs to match the search query. By default, partial matches are included in the results.</div>
+                        <label>Entire String</label>
```
TheoChevalier: Should be `<label for="entire_string">Each Word</label>`

app/views/search_form.php
```diff
@@ -168,8 +177,7 @@ class="search_options"
                                class="search_options"
                                <?=Utils::checkboxState($check['entire_words'])?>
                         />
-                        <label for="entire_words" aria-describedby="tooltip_entirewords">Entire Words</label>
-                        <div id="tooltip_entirewords" class="search_tooltip" role="tooltip">If selected, each search term needs to match an <strong>entire word</strong>. By default, partial word matches are included in the results.</div>
+                        <label>Entire Words</label>
```
TheoChevalier: Should be `<label for="entire_words">Entire Words</label>`

app/views/search_form.php
```diff
@@ -127,6 +127,18 @@ class="mainsearch_default_checkbox"
 
                 <fieldset id="advanced_search">
                     <label>Advanced Search options</label>
+                    <a href="#" id="tooltip_search"><span>i</span></a>
+                    <div class="tooltip" id="tooltip_search_text">
+                        <a href="#" id="tooltip_search_close" data-tooltip-id="tooltip_search_text" title="Close this panel">x</a>
```
TheoChevalier: nit: “×” character (U+00D7) vs. x (U+0078) for consistency with the clear button in the search field. If we’re switching characters, we will need to increase font-size

web/style/transvision.css
```diff
@@ -782,6 +790,83 @@ input[type="checkbox"]:disabled + label {
...
+    right: 8px;
+    top: 8px;
+    font-family: 'Fira Sans UltraLight', Arial, sans-serif;
+    font-size: 25px;
+    line-height: 0.5em;
+    color: #fff;
+}
+
+#tooltip_search_close:hover {
+    border: 1px solid #ccc;
```
TheoChevalier: Do we really need this? It feels a bit weird as it’s moving the button by 1px on hover. If we really want a border to identify the click area, maybe we should use a transparent border on the default state.

web/style/transvision.css
```diff
@@ -782,6 +790,83 @@ input[type="checkbox"]:disabled + label {
...
+
+#tooltip_search_close {
+    display: block;
+    position: absolute;
+    right: 8px;
+    top: 8px;
+    font-family: 'Fira Sans UltraLight', Arial, sans-serif;
+    font-size: 25px;
+    line-height: 0.5em;
+    color: #fff;
```
TheoChevalier: I think we want to add `text-decoration: none;` here to remove the underline from the link on hover

TheoChevalier:
I like this direction better than the current master (been annoyed by the tooltips too after a few days of usage). And I also think the always visible “i” icon won’t hurt for this critical area

app/views/search_form.php
flodolo: Gah, removed too much 😒

flodolo:
I think I've addressed all the issues, thanks for spotting them.

I removed the hover on the x, and slightly enlarged/moved it. Not sure why I didn't notice the text-decoration on hover :-\

flodolo:
Let me check that X again, it still doesn't look right.

flodolo:
> Let me check that X again, it still doesn't look right.

Ehm, never mind. VM didn't pull the latest changes…

PR #843, opened 2017-01-30
Track webcompat-reporter for Desktop

TheoChevalier:
Nothing critical, tracking the newly added folder. Do you think we *need* to wait for it to ride the trains, or it’s ok to merge during this cycle?


flodolo:
I noticed it too, I was mostly wondering about x-channel and if we're going to have it soon enough.

Anyhow, it won't hurt to fix this.

PR #844, opened 2017-01-31
Update logo with new Mozilla style

flodolo:


flodolo:
Per IRC discussion: we might want to revise this when the font is available, but this should be a decent start.

TheoChevalier:
Thanks! 🎉  However we probably don’t want full-white background on the PNGs?

flodolo:
Yeah, I fixed it. No idea why the SVG/PNG export adds background (I manually fixed the SVG in editor, for the PNG had to use another software to export)

flodolo:
Optimized PNGs with pngcrush too.

PR #846, opened 2017-02-01
Add md5 of repositories to TMX filename (fixes #841)

### LINKED ISSUE #841: TMX download broken when trying to download two different products
I set the TMX download page to en-US, Normal, Central and download the TMX.

If I then set the same option but for mozilla.org, I get the same TMX generated the first time instead of the new one.

I haven't used the download page in a while, not sure when it broke.

flodolo:


flodolo:
Looks like Bugzilla is having some issues, the usual query in mozilla-l10n-query failed last night, and it failed for me to in the browser.

Let's see if Travis passes this time.

app/models/tmx_downloading.php
```diff
@@ -47,16 +49,13 @@ function ($e) {
...
-    $content = TMX::create($strings, $locale, 'en-US');
-    $target_file = 'mozilla_en-US_' . $locale . '.tmx';
-} else {
-    $content = TMX::createOmegat($strings, $locale, 'en-US');
-    $target_file = 'mozilla_en-US_' . $locale . '_forOmegaT.tmx';
-}
-
-$target_file_path = WEB_ROOT . 'download/' . $target_file;
+$tmx_id = md5($requested_repositories);
+$target_file_name = "mozilla_en-US_{$locale}_{$tmx_id}_{$check['tmx_format']}.tmx";
```
TheoChevalier: Wondering if this is introducing a XSS vector, are we making sure `$check['tmx_format']` can’t contain anything else than "normal" or "omegat"?

flodolo: We are: https://github.com/mozfr/transvision/blob/master/app/inc/tmx_options.php#L4-L10

TheoChevalier: Good! Thanks for checking

PR #849, opened 2017-02-01
 Improve score on http://observatory.mozilla.org/ - fixes #808

### LINKED ISSUE #808: Improve score on http://observatory.mozilla.org/
We currently have a F, we may not be able to have a A+ without profound changes but we could easily target a B

TheoChevalier:
Based on what Pascal added to other mozFr websites.

### DIFF
--- a/app/views/templates/base.php
+++ b/app/views/templates/base.php
@@ -84,6 +84,11 @@ if (file_exists(CACHE_PATH . 'lastdataupdate.txt')) {
     print $page_title . ' | ';
 } ?><?= $title_productname ?></title>
     <meta charset="utf-8" />
+
+    <?php if (! LOCAL_DEV): ?>
+      <meta http-equiv="Content-Security-Policy" content="default-src https:; script-src https: 'unsafe-inline'; style-src https: 'unsafe-inline'; frame-ancestors 'none'">
+    <?php endif; ?>
+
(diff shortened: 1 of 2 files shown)


flodolo:
This breaks Transvision served via http, e.g. my local VM :-\

Not completely sure I want to force everyone to use a secure connection for local testing.

TheoChevalier:
hm, good point. And obviously we can’t add a condition on dev=true here…

flodolo:
I think it's possible to add the content policy as meta tag instead of htaccess?
https://developers.google.com/web/fundamentals/security/csp/#the_meta_tag

Not really familiar with the subject though.

TheoChevalier:
Unfortunately, not very familiar either :/ But I’ve updated the patch with your findings, seem to work

flodolo:
It works locally now, but it breaks a lot of stuff (tried to load it on my VM, which is using https).
* It's not loading inline SVG (e.g. i in search tooltip, swap locales, etc.). I guess we can switch to real .svg files
* We have inline scripts, for example the one determining which locales to display based on the channel or product. That doesn't work anymore, and I'm not sure we want to find a different solution.

flodolo:
What do we want to do with this one? Clearing r? for now

TheoChevalier:
Let’s close for now, I don’t have an easy way to fix this :/

PR #851, opened 2017-02-02
Changelog for v5.2

TheoChevalier:
I wonder if it’s worth keeping the “Webcompat Reporter” tracking or not

### DIFF
--- a/app/models/changelog_data.php
+++ b/app/models/changelog_data.php
@@ -65,18 +65,11 @@ $releases = [
 $changelog = [
   '5.2' => [
       [
-          'message' => ['Added link to Pootle or Pontoon under strings from the Aurora channel, in order to quickly edit translations'],
+          'message' => ['For locales working on Pootle or Pontoon, added links under Aurora search results to quickly access translations in the tool'],
           'section' => ['end_user'],
           'type'    => ['new'],
           'commit'  => ['6876a6d910f94b7f645ef3346aa3911ea2e37771'],
           'authors' => ['SkySymbol'],
-      ],
-      [
-          'message' => ['Added L20N support by switching to compare-locales for parsing product repositories'],
-          'section' => ['end_user'],
-          'type'    => ['new'],
-          'commit'  => ['6a4994dd79fbe93ddbd155792669c72a88d20ff2'],
-          'authors' => ['flod'],
       ],
       [
           'message' => ['Tracking Webcompat Reporter strings for Desktop'],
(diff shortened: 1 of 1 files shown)


app/models/changelog_data.php
```diff
@@ -59,9 +59,97 @@
...
+  '5.2' => [
+      [
+          'message' => ['Added link to Pootle or Pontoon under strings from the Aurora channel, in order to quickly edit translations'],
+          'section' => ['end_user'],
+          'type'    => ['new'],
+          'commit'  => ['6876a6d910f94b7f645ef3346aa3911ea2e37771'],
+          'authors' => ['SkySymbol'],
+      ],
+      [
+          'message' => ['Added L20N support by switching to compare-locales for parsing product repositories'],
```
flodolo: This is wrong, compare-locales doesn't support l20n (yet). I would say it's `code`, and note "Switched from Silme to compare-locales as library for parsing localization files in Mozilla products".

app/models/changelog_data.php
```diff
@@ -59,9 +59,97 @@
     '4.4'   => '2016-06-02',
     '5.0'   => '2016-08-04',
     '5.1'   => '2016-11-04',
+    '5.2'   => '2017-02-02',
 ];
 
 $changelog = [
+  '5.2' => [
+      [
+          'message' => ['Added link to Pootle or Pontoon under strings from the Aurora channel, in order to quickly edit translations'],
```
flodolo: Maybe: "For locales working on Pootle or Pontoon, added links under Aurora search results to quickly access translations in the tool."

app/models/changelog_data.php
```diff
@@ -59,9 +59,97 @@
...
+      ],
+      [
+          'message' => ['Added L20N support by switching to compare-locales for parsing product repositories'],
+          'section' => ['end_user'],
+          'type'    => ['new'],
+          'commit'  => ['6a4994dd79fbe93ddbd155792669c72a88d20ff2'],
+          'authors' => ['flod'],
+      ],
+      [
+          'message' => ['Tracking Webcompat Reporter strings for Desktop'],
```
flodolo: I would keep it out, since it's not available in Aurora?

app/models/changelog_data.php
```diff
@@ -59,9 +59,97 @@
...
+      ],
+      [
+          'message' => ['Fixed style issues on iOS and Safari'],
+          'section' => ['end_user'],
+          'type'    => ['bug'],
+          'commit'  => ['278fdc4468dda4f47fc45fb17ef6e4b1d7475d3c'],
+          'authors' => ['tchevalier'],
+      ],
+      [
+          'message' => ['“@@missing@@” could be seen under particular circumstances when strings are missing'],
```
flodolo: strings ➡️ reference/source strings

TheoChevalier:
Makes sense, updated :)

flodolo:
Looks good, thanks.

PR #852, opened 2017-02-02
Changelog: Use monospace font for commit hashes, drop 'commit'

flodolo:
Some small nits:
* Use a monospace font for the commit hash, so the description starts at the same point.
* Drop ‘commit’ in the link
* Don’t use spaces to separate commits or text, use CSS margin
* Don’t show a nickname on a new line, all alone.

It might be overkill to include the FiraMono font, but we might use it elsewhere. In that case we would just need to move it to the main CSS.


flodolo:
This thing bothers me more that it should

schermata 2017-02-02 alle 19 12 51

flodolo:
Post intervention

schermata 2017-02-02 alle 19 18 11

TheoChevalier:
Looks great, kudos for loading the font on this page only ;)

Issue #822, opened 2016-11-06
Results should be more visible after permorming a search


flodolo:
Reported by Julen at the Paris hackathon.

After searching you end up in the results page, but you have to scroll down a bit to get the results. Check google.com for example.

TheoChevalier:
In addition to what will land in #853, once x-channel has landed, I wonder if we can remove the main title “You are looking at the xxx channel”, and reposition the locale code elsewhere. I think the channel info will be less critical, and maybe we can merge it with “Displaying xx results for the string xx in ab-CD:”?

flodolo:
I'll leave this one open. And yes, I think we can remove it.

flodolo:
I think we can close this one.

PR #853, opened 2017-02-03
Make search results more visible

flodolo:


flodolo:
This should be a usable starting point for #822:
* Move API promotion to the bottom the page. Given the target, it doesn't seem suitable from a high visibility area. 
* Jump to the filters if they're available, and there's no other hash defined in the URL. Also add filters to entities result

flodolo:
The only issue is when the page is really heavy and long, the animation is a big laggy, but there's not much we can do.

PR #860, opened 2017-03-01
Fix Pootle/Pontoon edit link for products other than Firefox (fixes #859)

### LINKED ISSUE #859: Pontoon and Pootle links for mail, suite and mobile are wrong
We always point to the Firefox project, we should differentiate them.

flodolo:
fixes #859


flodolo:
Thanks.

PR #861, opened 2017-03-05
Update php-cs to 2.x

flodolo:
I think I've ported all existing rules, plus added a bunch for whitelines and phpdocs

Too bad you can't check for `! $var`, only the other way around (make sure there's no space)..

Also moved to .php_cs.dist, as documentation suggest
https://github.com/FriendsOfPHP/PHP-CS-Fixer

Upgrade doc: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/UPGRADE.md

It looks like we should use lowercase to describe params (it does that online for multiline comments)


flodolo:
Thoughts? For example we could remove `phpdoc_annotation_without_dot` and undo the changes it made, which are a bit weird unless we go in and fix the comments.

flodolo:
Actually, I gave it a try without the `phpdoc_annotation_without_dot` rule (which is confusing with the change of case).

Added `single_quotes` on the other hand, which seems really useful.

TheoChevalier:
Looks good 👍  The `single_quote` thing is smart when it’s mixed with double quotes, so that’s great, like here https://github.com/mozfr/transvision/pull/861/commits/c8504ca7fba5cb3e4c96c6b5013ea41a34e6cad6#diff-97e1b28d9e9c2cd94bad73816bf7aca8L65
And I’m good with removing `phpdoc_annotation_without_dot`

PR #863, opened 2017-03-19
Fix #862 - Add the link to Pootle / Pontoon on the entity view

### LINKED ISSUE #862: Edit in Pontoon|Pootle should be available also in Entities search
Right now it's not available in that view.

SkySymbol:


app/views/results_entities.php
```diff
@@ -191,6 +198,7 @@
       <div dir='ltr' class='result_meta_link'>
         <a class='source_link' href='{$path_locale2}'>&lt;source&gt;</a>
         {$file_bug}
+        {$edit_link}
```
flodolo: You should use the same order of the other view (source, edit, bug from left to right).

SkySymbol: Done :-)

PR #864, opened 2017-04-14
Remove Aurora channel

flodolo:


flodolo:
Pending on https://github.com/mozilla-l10n/mozilla-l10n-query/pull/53

Tests will fail, you can test locally by using https://test.flod.org/mozilla-l10n-query/

flodolo:
Tests are green, it should be easy to test it.

PR #869, opened 2017-04-24
Fixes #865 - Disable the Search button if the input box is empty

### LINKED ISSUE #865: Empty Query is also process for search
If **Search Query** box is empty then it must not process the _request_, but its not **validating** the query box whether it is **empty or not**.

SkySymbol:


TheoChevalier:
This is quite a sensitive area, I’d suggest we rely on the browser to do the check for us, why not just adding "required" to the input field? I think it covers what we need 😉

I’ve also noticed that the button remains disabled if you try to run the same search (search something, then click again on search)

SkySymbol:
Right, this is a much simpler solution

PR #875, opened 2017-06-12
Fixes #866 - Sanitize strings that will be displayed in the unchanged view

### LINKED ISSUE #866: Clicking on Menu button on specific page reloads the same page
visiting this page

https://transvision.mozfr.org/unchanged/
https://transvision.mozfr.org/productization/

and on click Menu, it reloads the same page, href problem

SkySymbol:


flodolo:
While this works, it's also highly inefficient. Instead of applying this to the whole array, use secureText on the string before displaying it
https://github.com/mozfr/transvision/blob/master/app/views/unchanged_strings.php#L46-L47

Example: French page 27 MB 0,19s with this code, 20MB and 0,08 with the suggested changes. For en-US memory is 55 MB vs 64 MB.

SkySymbol:
Ok, I though it was better to call the function once per array. Thanks!

PR #876, opened 2017-06-16
Ignore en-US in some views

flodolo:


flodolo:
Review ping when you're around :-)

TheoChevalier:
Sorry for the delay, traveling for the next two weeks so will try to find time but can’t guarantee

TheoChevalier:
Hm, seems like loading `http://localhost:8082/unchanged/?locale=en-US&repo=central` displays `ach` in the dropdown list (as expected), but still displays the en-US content? `25469 strings identical to English`
I assume we want to fallback to the default value of $locale if there is an ignored locale in the URL? Like I fall back for `fr` if I try to load `http://localhost:8082/unchanged/?locale=foobar&repo=central`

flodolo:
> I assume we want to fallback to the default value of $locale if there is an ignored locale in the URL?

Good point, but we actually fall back to the first language in the list, not the default one
https://github.com/mozfr/transvision/pull/876/commits/e479ba7a82f7587dda2c669c2e3a8a224985f333

You fall back to the default one is the locale doesn't exist in the list of supported locales.

flodolo:
Side note: we might want to work on improving that logic (also, display somewhere that the locale requested is not available and we're falling back to something else), but that requires a lot more code.

flodolo:
Also figured out how to make tests pass on Trusty.

Issue #883, opened 2017-08-27
Add links to Pontoon from the Access Keys report


bidaian:
The [QA page for checking access keys](https://transvision.mozfr.org/accesskeys/) lists the problems but does not link them back to Transvision or Pontoon, so to fix them you have to manually look for them.

flodolo:
I think it would make sense to include this in a much larger rewrite, which includes using only one piece of code to write results.

Right now they're scattered all over the place, resulting in these differences between views. It also means this would become a quite big change.

flodolo:
This was fixed.

PR #884, opened 2017-09-09
Add Pontoon links to QA views, improve accesskeys view

### LINKED ISSUE #870: Add the handy links to the pages in the QA View
When I use Access Keys and Translation Consistency views really miss the direct link to translation tool but suppose others can benefit from having them in the other views too.

flodolo:
Fixes #870 and #883. It's a first draft (seems to be working), but not ready for review yet.

The change to getEditLink is needed because I added a pontoon-mozorg to mozilla-l10n-query yesterday (for the webdashboard), since we have locales using Pontoon for mozilla.org but not for products (e.g. it, da, de, etc.).

Result: now we're showing Pontoon links to every locale.

### DIFF
--- a/app/models/accesskeys.php
+++ b/app/models/accesskeys.php
@@ -33,17 +33,30 @@ $target = Utils::getRepoStrings($locale, $repo);
 $ak_string_ids = array_filter(
     array_keys($target),
     function ($entity) {
-        return strpos($entity, 'accesskey') !== false;
+        return strpos($entity, '.accesskey') !== false;
     }
 );
 
 // Possible labels associated to an access key
-$ak_labels = ['label', 'title', 'message'];
+$ak_labels = ['.label', '.title', '.message', ''];
+
+// Known false positives
+$ignored_ids = [
+    'suite/chrome/mailnews/messenger.dtd:searchButton.title',
+];
 
 $ak_results = [];
 foreach ($ak_string_ids as $ak_string_id) {
     foreach ($ak_labels as $ak_label) {
-        $entity = str_replace('accesskey', $ak_label, $ak_string_id);
+        /*
+            Replace 'accesskey' with one of the known IDs used for labels.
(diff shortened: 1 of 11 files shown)


flodolo:
It should be ready for review.

flodolo:
The accesskey logic was quite poor: 

1.  [Check](https://github.com/mozfr/transvision/blob/master/app/views/accesskeys.php#L24-L29) if the last 9 characters of a string ID are `accesskey`.
2. Remove 10 characters, add one of `['.label', '.title', '.title2']` and see if such a string ID exists.

I've already improve that logic on step 1, by searching for `accesskey` in the entire string (that would catch `accesskey2` for example), but I've realize we miss out on all DevTools access keys, where the schema is often:

```
foo = Some label
foo.accesskey = S
```

I need to think a bit more about this.

flodolo:
Improved logic.

Esperanto moves from 17 to 24 access keys, there's even one for French.

app/models/accesskeys.php
```diff
@@ -0,0 +1,96 @@
...
+            if (in_array($entity, $ignored_ids)) {
+                continue;
+            }
+            /*
+                Store the string if the access key is empty or using a
+                character not available in the label.
+            */
+            if ($current_ak == '') {
+                $ak_results[$ak_string_id] = $entity;
+            } elseif (mb_stripos($target[$entity], $current_ak) === false) {
```
TheoChevalier: I assume it’s to avoid a (too) long condition in one `if`?

flodolo: No, this slipped through the various rewrites.

flodolo:
> dir/lang attributes are missing on <div class="result_meta_link">, so for RTL locales, the <source> and <Pontoon> links are not displayed properly

Ugh, good catch. Fixed also the other view where I copied the code

PR #893, opened 2017-09-18
(fixes #892) Remove hashes from URL when submitting simple search forms

### LINKED ISSUE #892: Remove hashes from URL when using simple search form
Noticed another strange behavior while testing the views with components filter.

If, for any reason, I end up with the hashtag in the URL
https://transvision.mozfr.org/variables/#mobile

Submitting the form, e.g. changing the locale, will get you to
https://transvision.mozfr.org/variables/?locale=fr&repo=central#mobile

Which doesn't make a lot of sense. We should strip the hashtag when possible.

flodolo:


flodolo:
OK, this is not working in all cases. Checking.

flodolo:
OK, should be good to review.

PR #895, opened 2017-09-24
Implement cross-channel localization support, add Focus for iOS and Android

flodolo:
Also:
* Remove 'current' title from main search
* Udpate to compare-locales 2.1 for parsing
* Add Focus for iOS
* Drop channel comparison view


flodolo:
For testing, I suggest to rename the existing data folder. You'll need to point the mozilla-l10n-query to https://test.flod.org/mozilla-l10n-query/

Before merging:
* I'll need to fix a few TODOs (e.g., I should point to the gecko-strings, not gecko-strings-quarantine).
* I'll need to disable the hook on my VPS (if we want to keep release there).

Changes to the tmx_maker script are covered here
https://github.com/flodolo/tmx_maker/pull/7

flodolo:
I'm trying to change tests so that they don't fail in PRs like this.

TheoChevalier:
For the sake of testing, do we have any string that has been updated in 58 and not in 57? If not I guess I can fake it locally, but I’d prefer using a real one.
In such cases we should have both versions in the repo, correct?

flodolo:
> In such cases we should have both versions in the repo, correct?

You mean string change without ID? In that case, the repo would only have one version (from last update, likely from mozilla-central).

For now, the user repo only has 57.

TheoChevalier:
No, I wanted to test the following case:
- Assuming the repo is used to build all branches
- Version a of the string is used for beta
- Version b (with a different ID) is used for central

We won’t have a test repo with this kind of data until x-channel is fully ready?

flodolo:
> We won’t have a test repo with this kind of data until x-channel is fully ready?

We should have one next week.

TheoChevalier:
Okay, nice. I have plenty of things to test and review until then anyway

flodolo:
> Okay, nice. I have plenty of things to test and review until then anyway

How do you feel if I add the work I've been doing on top of this to this PR for po files?
https://github.com/flodolo/transvision/compare/x-channel...flodolo:po

It's a pretty drastic change, including:
* Add Focus for Android and a .po parser
* Reorganize project information, removing duplication
* Use md5 to generate IDs for mozilla.org, focus, etc. (will break existing links), change the way XLIFF IDs are generated
* Fix pattern checks

In case I need to add Focus for Android to mozilla-l10n-query.

flodolo:
Updated PR, also updated the test mozilla-l10n-query instance

TheoChevalier:
I wonder, how hard/complex would it be to redirect links from `&repo=central` (or beta/release), to `&repo=gecko-strings`? I’d like to avoid breaking existing links, if it’s not too complicated or doesn’t add too much complexity to our code. Like, if we just need basic mapping to `$repos_lists` and `$repos_infos`, is it acceptable?

Alternatively, or in addition to that, we could replace the current (somewhat rude) error `Entity does not exist for this repo, goodbye.` with something more helpful, like:

`This repository does not exist, would you like to try to <a>perform this search on the Gecko string repository?</a>`

flodolo:
That already happens for me: if you search, you get gecko_strings automatically. The problem it's only when you access directly the onestring view.

TheoChevalier:
Right, this happened for the onestring view. Unfortunately that’s the one that is the most popular in Bugzilla or GitHub comments :/

flodolo:
> Right, this happened for the onestring view. Unfortunately that’s the one that is the most popular in Bugzilla or GitHub comments :/

That's actually easy to fix (and it was one of the pieces of code I missed).

flodolo:
Looks like I broke something in Health View.

TheoChevalier:
Running setup.sh again creates a merge `Merge branch 'x-channel' of https://github.com/flodolo/p12n-extract into x-channel`, not sure why yet

flodolo:
> Running setup.sh again creates a merge Merge branch 'x-channel' of https://github.com/flodolo/p12n-extract into x-channel, not sure why yet

That's weird. Even removing the folder and trying to run setup twice? It's only pulling and checking out the branch (which will be removed before merging this).

TheoChevalier:
> Even removing the folder and trying to run setup twice?

This works, so let’s pretend nothing happened :)

flodolo:
> That's weird. Even removing the folder and trying to run setup twice? It's only pulling and checking out the branch (which will be removed before merging this).

I know why… I had to force push for a rebase the other day against master, if you already the folder obviously it didn't like that.

TheoChevalier:
https://pastebin.mozilla.org/9069414
This happened when running glossaire.sh without arguments, when it actually updated repositories.
But I can’t reproduce, even with force-tmx…

flodolo:
It looks like it's not founding the Gettext\Translations class, which is quite weird if it works otherwise.
In case, it's the tmx_gettext script, and it runs independently for the force-tmx flag

flodolo:
Added a tweak to improve behavior and logging in GitHub webhook. For some reasons it still fails on my VPS (it completes, but GH thinks it timed out), but I think it's worth taking the improvement.

TheoChevalier:
I always have shivers when I see changes to webhooks, but this looks safe-ish. Are we using those headers in other places already?

flodolo:
> I always have shivers when I see changes to webhooks, but this looks safe-ish. Are we using those headers in other places already?

Well, we only use the webhook on beta. A very similar code is used on Webstatus, and it works fine, so I suspect it's something related to nginx on my VPS.

flodolo:
> Should we pick a date for putting this one out?

Worth talking with @Pike, since he plans to move the gecko-strings repository to the official place.

Things to do before merging this one:
1. Fix TODOs (depends on Pike's answer)
2. Disable webhook on Beta
3. ~~Disable webhook on my VPS~~

Things before releasing to prod:
1. Stop cronjobs on the server
2. Merge dependencies from external repositories
3. Merge, rename data folders for both Beta and production, run setup and glossaire

flodolo:
I've removed the TODOs. Note that I can't disable the webhook, I'm not admin of this repo ;-)

TheoChevalier:
Webhook disabled!

PR #896, opened 2017-10-07
Display filtering info for gecko_strings only in Consistency view

TheoChevalier:
Noticed this while reviewing x-channel, will rebase once x-channel has landed.

To reproduce, check consistency for SeaMonkey for instance, then check consistency for Firefox for iOS. It’s unrelated to the x-channel patch.

Also removing some Experimental flags, switching to `unlocalized` by default and preventing huge log files on the server due to Python process crashes.


TheoChevalier:
Don’t have much more, should be ready for review (the last 3 patches, obviously)

flodolo:
Uhm, now I think of it, it's going to be super painful.

I still need to fix the TODOs before merge. Once we merge, we squash the commits, so git is going to have a hell of a time in rebasing against master. You can try, but it's likely easier to create a new branch post merge with the fixes in one commit.

TheoChevalier:
I wonder if manually squashing all the x-channel commit in this branch only would do the trick and make it mergeable once x-channel lands

TheoChevalier:
Or, with minor conflicts with the TODOs

flodolo:
> I wonder if manually squashing all the x-channel commit in this branch only would do the trick and make it mergeable once x-channel lands

Not sure, probably it would make it at least less painful since you might get stuck on each commit.

app/scripts/setup.sh
```diff
@@ -63,10 +63,14 @@ function setupExternalLibraries() {
         echogreen "Checking out the p12n-extract library in $libraries"
         cd $libraries
         git clone https://github.com/flodolo/p12n-extract/ p12n
+        #TODO: remove this
+        cd p12n;git checkout x-channel;cd ..
```
flodolo: There should be no changes to setup.sh

TheoChevalier: hm, looks like GitHub decided it was fine to include those… :/

README.md
```diff
@@ -3,7 +3,7 @@
...
 
-The main purpose of Transvision is to provide a specialized search engine to find localized strings in Mozilla code repositories for all Mozilla products (Firefox, Thunderbird, Seamonkey) and websites (currently only www.mozilla.org is supported) via a Web interface. There are also side-features such as checks for common typographical errors for some languages, validity checks for localized access keys in the UI, or comparison views between Mozilla repository channels (Nightly/Beta/Release).
+The main purpose of Transvision is to provide a specialized search engine to find localized strings in Mozilla code repositories for all Mozilla products (Firefox, Firefox for Android, Firefox for iOS, Focus for iOS, Focus for Android, Thunderbird, SeaMonkey) and websites (currently only www.mozilla.org is supported) via a Web interface. There are also side-features such as checks for common typographical errors for some languages, validity checks for localized access keys in the UI, or comparison views between Mozilla repository channels (Nightly/Beta/Release).
```
flodolo: Need to remove the last sentence here (comparison)

PR #903, opened 2017-10-14
Fix Gettext and XLIFF parsers

flodolo:
Fixing a couple of issues/nits found while working on a different project.


TheoChevalier:
Maybe it’s worth editing the test file https://github.com/mozfr/transvision/blob/master/tests/testfiles/xliff/firefox-ios.xliff and add the same ID in two different file sections?

flodolo:
> Maybe it’s worth editing the test file https://github.com/mozfr/transvision/blob/master/tests/testfiles/xliff/firefox-ios.xliff and add the same ID in two different file sections?

Good call, adding.

PR #908, opened 2017-10-18
(fix #907) Trim query when searching for entities

### LINKED ISSUE #907: Trim query when searching for entities
If you [search for strings or strings+entities](https://transvision.mozfr.org/?recherche=+test&repo=gecko_strings&sourcelocale=en-US&locale=it&search_type=strings), and have extra spaces in the query, you get a warning and a link to search without spaces.

That doesn't happen on the "entities" search, and that's bad, since a search with spaces will basically always fail for entities.

We should trim the search query.

flodolo:
Also clean up code, using class method instead of a variable defined in a random include.

### DIFF
--- a/app/controllers/mainsearch.php
+++ b/app/controllers/mainsearch.php
@@ -68,12 +68,12 @@ if ($check['t2t']) {
     require_once VIEWS . 'results_glossary.php';
 } else {
     // No search
-    if ($my_search == '') {
+    if ($search->getSearchTerms() == '') {
         return;
     }
 
     // Search not acceptable
-    if (mb_strlen(trim($my_search)) < 2) {
+    if (mb_strlen($search->getSearchTerms()) < 2) {
         print '<p><strong>Search term should be at least 2 characters long.</strong></p>';
 
         return;
--- a/app/controllers/mainsearch.php
+++ b/app/controllers/mainsearch.php
@@ -54,6 +54,17 @@ if ($url['path'] == '3locales') {
     require_once MODELS . '3locales_search.php';
 }
 
+/*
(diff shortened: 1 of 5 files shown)


TheoChevalier:
Since we’re actively trimming the query (compared to String search where we just warn), we may want to update the value in the search field. Just a suggestion, it’s looking good as it is

app/controllers/mainsearch.php
```diff
@@ -54,6 +54,17 @@
...
 }
 
+/*
+    Trim the search query when searching for entities, display a warning.
+    This needs to be done before including search_form.php, in order to
+    display the trimmed value in the search field.
+*/
+if ($search->getSearchType() == 'entities' &&
+    $search->getSearchTerms() !== trim($search->getSearchTerms())) {
+    $search->setSearchTerms(trim($search->getSearchTerms()));
```
flodolo: I guess Atom's autocomplete here failed me, and it was using `setRegexSearchTerms` before.

app/views/results_entities.php
```diff
@@ -225,7 +225,7 @@
 if ($entities) {
     $message_count = $real_search_results > $limit_results
         ? "<span class=\"results_count_{$search_id}\">{$limit_results} results</span> out of {$real_search_results}"
-        : "<span class=\"results_count_{$search_id}\">" . Utils::pluralize(count($search_results), 'result') . '</span>';
+        : "<span class=\"results_count_{$search_id}\">" . Utils::pluralize(count($real_search_results), 'result') . '</span>';
```
flodolo: Poor copy and paste: `$search_results` is not defined in this view.

PR #916, opened 2017-11-05
Expose latest tag name in the footer and automatically generate a rel…

TheoChevalier:
…ease in /news

Didn’t implement the same changes to the RSS feed, because:

1/ I think it doesn’t really make sense to publish temporary things there (it will disappear at the next documented release)
2/ I’m lazy 😇

### DIFF
--- a/app/models/changelog.php
+++ b/app/models/changelog.php
@@ -62,7 +62,7 @@ $get_sections = function ($section) {
 };
 
 // Helper to generate the list of patches for the version
-$github_link = function ($release) use ($releases) {
+$github_link = function ($release, $releases) {
     if ($release > 1) {
         $keys = array_keys($releases);
         $previous = $keys[array_search($release, $keys, true) - 1];
@@ -76,7 +76,7 @@ LINK;
 };
 
 // Helper to generate a release title block
-$release_title = function ($version) use ($releases) {
+$release_title = function ($version, $releases) {
     return <<<TITLE
 <h2 class="release_number" id="v{$version}"><a href="#v{$version}">Version {$version}
 <span class="release_date">{$releases[$version]}</span></a>
@@ -121,4 +121,21 @@ $authors = function ($list) use ($author_urls) {
     }
 
     return $text;
+};
+
+$latest_undocumented_release = function() use ($releases, $github_link, $release_title) {
+  $text = '';
+  $version_number = str_replace("\n", '', str_replace('v', '', file_get_contents(CACHE_PATH . 'tag.txt')));
+  end($releases);
+  $last_release = key($releases);
+
+  if ($version_number != $last_release) {
+    $github_relnotes = "https://github.com/mozfr/transvision/releases/tag/v{$version_number}";
+    $releases[$version_number] = 'Unknown';
+    $text .= $release_title($version_number, $releases);
+    $text .= "<p>This release is undocumented, but you can know more in the <a href=\"{$github_relnotes}\">GitHub release notes</a> and with the list of commits below since the last documented release.</p>";
+    $text .= $github_link($version_number, $releases);
+  }
+
+  return $text;
 };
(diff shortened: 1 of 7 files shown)


app/scripts/setup.sh
```diff
@@ -161,16 +161,19 @@ DEV_VERSION="dev"
...
     CURRENT_TIP=$(git rev-parse HEAD)
     LATEST_TAG=$(git describe --abbrev=0 --tags | xargs -I {} git rev-list -n 1 {})
+    LATEST_TAG_NAME=$(git describe --abbrev=0 --tags)
     if [ "${CURRENT_TIP}" = "${LATEST_TAG}" ]
     then
         DEV_VERSION=""
     fi
 fi
 echo "${CURRENT_TIP:0:7}${DEV_VERSION}" > "${install}/cache/version.txt"
+echo "${LATEST_TAG_NAME}" | tr -d '\n' > "${install}/cache/tag.txt"
```
flodolo: I'd keep the new line here, and deal with it when reading the file like we do for `version.txt`.

```
echo "${LATEST_TAG_NAME}" > "${install}/cache/tag.txt"
```

app/views/templates/base.php
```diff
@@ -75,6 +75,11 @@
 } else {
     $last_update = "<p>Data last updated: not available.</p>\n";
 }
+
+$version_number = '';
+if (file_exists(CACHE_PATH . 'tag.txt')) {
+    $version_number = $title_productname . ' ' . file_get_contents(CACHE_PATH . 'tag.txt') . "\n";
```
flodolo:
```
$version_number = $title_productname . ' ' . file(CACHE_PATH . 'tag.txt', FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
```

I don't think we need the new line at the end.

app/models/changelog.php
```diff
@@ -122,3 +122,20 @@
...
+  $text = '';
+  $version_number = str_replace("\n", '', str_replace('v', '', file_get_contents(CACHE_PATH . 'tag.txt')));
+  end($releases);
+  $last_release = key($releases);
+
+  if ($version_number != $last_release) {
+    $github_relnotes = "https://github.com/mozfr/transvision/releases/tag/v{$version_number}";
+    $releases[$version_number] = 'Unknown';
+    $text .= $release_title($version_number, $releases);
+    $text .= "<p>This release is undocumented, but you can know more in the <a href=\"{$github_relnotes}\">GitHub release notes</a> and with the list of commits below since the last documented release.</p>";
```
flodolo: Maybe: There is no entry in the changelog for this release, but you can learn more about it in the GitHub release notes and look at the list of commits from the latest documented release.

app/models/changelog.php
```diff
@@ -122,3 +122,20 @@
 
     return $text;
 };
+
+$latest_undocumented_release = function() use ($releases, $github_link, $release_title) {
+  $text = '';
+  $version_number = str_replace("\n", '', str_replace('v', '', file_get_contents(CACHE_PATH . 'tag.txt')));
```
flodolo: No need to replace `\n`,  read the file with `FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES` flags. 
I would also just strip the first character from the tag.

app/scripts/setup.sh
TheoChevalier: We really don’t want new lines in this file, so it’s not obvious to me why it’s better to take the risk of forgetting removing it on the other end

TheoChevalier: Because if for any reason someone edits the file manually?

flodolo: I think it's mostly a matter of leaving the command output alone. If you prefer to remove the new line here (note that your code is already doing it twice though), I'd update the code for the version to follow the same logic.

PR #918, opened 2017-11-07
(fixes #917) Fix escaping in Pontoon links (entities search)

### LINKED ISSUE #917: Edit in Pontoon doesn't work with escaped HTML entities
Originally reported as https://bugzilla.mozilla.org/show_bug.cgi?id=1415084

flodolo:


TheoChevalier:
Looks like you need to update one of the tests ;)

TheoChevalier:
Actually, should `<` and `>` be excluded as well? It seems it doesn’t fix the case of `<br>`

flodolo:
Ugh, I think I'm trying to fix the wrong problem. I just realized that the edit link works just fine in the [main search](https://transvision.mozfr.org/?recherche=Pr%C3%A1v%C4%9B+na+tohle+jste+%C4%8Dekali.%3Cbr%3ENejrychlej%C5%A1%C3%AD+Firefox+v%C5%A1ech+dob.&repo=mozilla_org&sourcelocale=en-US&locale=cs&search_type=strings), but it fails for the [entity view](https://transvision.mozfr.org/?sourcelocale=en-US&locale=cs&repo=mozilla_org&search_type=entities&recherche=mozilla_org/firefox/whatsnew_57.lang:05cf0ffdba870415c4ebd9ec134caab4&entire_string=entire_string).

So the problem is not using `urlencode` at all.

flodolo:
Turns out we were running urlencode on the escaped version of the string.

app/views/results_entities.php
```diff
@@ -26,9 +26,10 @@
     $path_locale2 = VersionControl::getPath($locale, $current_repo, $entity);
 
     // Escape strings for HTML display
```
TheoChevalier: This comment probably need to be updated?

flodolo: Moving

Issue #897, opened 2017-10-07
Revisit logo font


TheoChevalier:
@flodolo Do you still have the .psd source for the current logo?

We may want to switch to Zilla slab highlighted for `Transvision`

flodolo:
I'm pretty sure I just edited the svg file directly
https://github.com/mozfr/transvision/pull/844

TheoChevalier:
Hm, you’re definitely old-school :) Ok, I will check how much I suck at Photoshop

TheoChevalier:
It looks great here, but I’m worried there would be too much blue with the current bluish background… What do you think about something like that, and without capitalizing Transvision? https://static1.squarespace.com/static/58879499725e2572b9c4e0ec/t/592dfe1eb3db2bf8af733d4a/1496186413997/?format=1500w

TheoChevalier:
But yeah, like this one much, much better! :)

PR #920, opened 2017-11-07
(fixes issue #897) Use Zilla font in Transvision

flodolo:
fixes issue #897


flodolo:
Black Zilla Highlight (don't like it very much)

schermata 2017-11-07 alle 18 05 47

flodolo:
Black highlight, white fill for the text

schermata 2017-11-07 alle 18 11 10

TheoChevalier:
Thanks for doing that, @flodolo! I think the latest version looks great, happy to ship that one if it works for you as well.
Also I can’t help but notice the mozilla logo is not *exacly* how it’s supposed to be (paddings are a bit off), do you want me to ask the designers in my team if they would accept to help us here?

flodolo:
> Also I can’t help but notice the mozilla logo is not exacly how it’s supposed to be (paddings are a bit off), do you want me to ask the designers in my team if they would accept to help us here?

The fun thing is that the padding is produced by the Zilla Slab Highlight font. I could do it manually, if you mean distributing the padding evenly around the text.

flodolo:
> do you want me to ask the designers in my team if they would accept to help us here

If they can, go for it. We only have the SVG file though.

flodolo:
Using Zilla and manually placing highlights around text

schermata 2017-11-08 alle 17 08 01

flodolo:
Interesting: the font is bold 46pt, and yet it's completely different between Zilla (heavier) and Zill Highlight.

TheoChevalier:
Thanks, it looks really nice! No need to bother the designers, then.
Just noticed we’re not using `$cache_bust` on the logo URL, do you think we should, or the default value for svg is already short enough?

flodolo:
> Just noticed we’re not using $cache_bust on the logo URL, do you think we should, or the default value for svg is already short enough?

Default is one month. Let me squash and add it.

PR #927, opened 2017-12-13
Improve FTL support

flodolo:
* Fix edit link with strings with attributes
* Add support for variables checks


flodolo:
@TheoChevalier 
If possible, I'd like to merge and push out to prod the coming week. We have one string, and locales already errors (translating the variable) :-\

We'll have another PR later for Fluent 0.5, but I think it's worth covering FTL in the meantime.

flodolo:
> Looks good! But I think we should add a test file with as many edge cases as we could for FTL

I agree. This adds a very basic support. I have the feeling that we need to improve as we go (i.e. as we add new strings). For example, I discovered only now that `Quit { $brandname }` becomes `Quit { brandname[accusative] }` if you call for a variant…

PR #928, opened 2017-12-14
Observatory

ohm1998:
i want to contribute.


flodolo:
Hi, thanks for your interest.

Having said that, I'm going to close this: you opened a PR from a dead branch to master without fully understanding what the problem is. The branch was dead because it turned out to be really hard to fix the issue: https://github.com/mozfr/transvision/pull/849

This requires a full installation of Transvision, with content served through Apache, and a good understanding of the entire system. Some issues might not even be fixable, so it's definitely not a good first bug to fix.

ohm1998:
Hi,
Thanx for prompt reply.

Regards,
Ohm

Issue #910, opened 2017-10-20
False positive accesskey error


MikkCZ:
For Czech there is a false positive accesskey error for `decoder.noCodecs.accesskey` (see https://transvision.mozfr.org/accesskeys/?locale=cs&repo=gecko_strings). But it should belong to `decoder.noCodecs.button` instead of `decoder.noCodecs.message`. I am not sure, how exactly the detection works, but can you prefer `.button` before `.message` suffix?

flodolo:
The logic is that we strip .accesskey (e.g. getting `decode.noCodecs`), the try one of 

```
$ak_labels = ['.label', '.title', '.message', ''];
```

At the moment we don't even look for `.button`. I wonder if adding this as an [exception](https://github.com/mozfr/transvision/blob/master/app/models/accesskeys.php#L44) would be better than trying to fit it into the logic.

PR #930, opened 2017-12-21
Add known mappings for access keys in sync preferences

MikkCZ:
These access keys have been added in [bug 1395453](https://bugzilla.mozilla.org/show_bug.cgi?id=1395453) ([changeset](https://hg.mozilla.org/l10n/gecko-strings/rev/743121cc47e3)) and I have noticed [FP for `engine.creditcards.accesskey`](https://transvision.mozfr.org/accesskeys/?locale=cs&repo=gecko_strings) today. Adding them all to avoid potential more FPs for other locales or in future.

Btw. in case there is a change in any of the IDs on both sides of the mapping, is there some reporting that the mapping is not valid? In general the approach of known mappings won't scale indefinitely and this can at least help to maintain the existing ones.

### DIFF
--- a/app/models/accesskeys.php
+++ b/app/models/accesskeys.php
@@ -48,8 +48,14 @@ $ignored_ids = [
 // Some keys map to a different string ID than the one identified by the
 // algorithm
 $known_mappings = [
-    'browser/chrome/browser/browser.properties:decoder.noCodecs.accesskey' => 'browser/chrome/browser/browser.properties:decoder.noCodecs.button',
-    'extensions/irc/chrome/pref-irc.dtd:pref-irc.open.accesskey'           => 'extensions/irc/chrome/pref-irc.dtd:pref-irc.open.label',
+    'browser/chrome/browser/browser.properties:decoder.noCodecs.accesskey'      => 'browser/chrome/browser/browser.properties:decoder.noCodecs.button',
+    'extensions/irc/chrome/pref-irc.dtd:pref-irc.open.accesskey'                => 'extensions/irc/chrome/pref-irc.dtd:pref-irc.open.label',
+    'browser/chrome/browser/preferences/sync.dtd:engine.tabs.accesskey'         => 'browser/chrome/browser/preferences/sync.dtd:engine.tabs.label2',
+    'browser/chrome/browser/preferences/sync.dtd:engine.logins.accesskey'       => 'browser/chrome/browser/preferences/sync.dtd:engine.logins.label',
+    'browser/chrome/browser/preferences/sync.dtd:engine.prefs.accesskey'        => 'browser/chrome/browser/preferences/sync.dtd:engine.prefs.label',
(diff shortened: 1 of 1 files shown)


flodolo:
> These access keys have been added in bug 1395453 (changeset) and I have noticed FP for engine.creditcards.accesskey today. Adding them all to avoid potential more FPs for other locales or in future.

The view is not going to catch all access keys, because naming of string IDs is quite inconsistent. We're aware of that, and also know that the problem is going to disappear with FTL.

The `$know_mappings` array is used to avoid false positives in the view, not to catch more accesskeys than the logic allows.

Having said all that, I'm quite lost on the linked changeset. Are we talking about `engine.tabs.`? Because it was already there, it's not a new string, and it's a known casualty of the naming mess.

> Btw. in case there is a change in any of the IDs on both sides of the mapping, is there some reporting that the mapping is not valid? In general the approach of known mappings won't scale indefinitely and this can at least help to maintain the existing ones.

Not at the moment. It can't be tested because it requires real live data, and throwing warnings on running instances seems less than ideal.

As said, that array's purpose is to avoid false positive, not to track down all exceptions, so I don't expect it to grow.

flodolo:
OK, I completely missed that this was a PR and not an issue 🦄 

I believe this should be closed: as explained, that array is to avoid know false positives coming out of the current logic. 

A solution I'd be open to would be to change the logic to search for variations of `$ak_label`(e.g. label2, message2), but I'm not completely sure it's worth the effort.

MikkCZ:
The specific FP I see for Czech is that `engine.addresses.accesskey` is being mapped to `engine.creditcards.title` instead of `engine.creditcards.label`. I have added this mapping the same way as you fixed #910 + for the rest where `.title` entity was added in the same changeset that introduced `engine.creditcards.title`.

I actually did not noticed `engine.tabs.label2` is `.label2` at all. :) It will produce false positives the same way anyway.

flodolo:
OK, this makes a lot more sense now, sorry for the confusion. I assumed you were trying to catch access keys not caught by the logic, not adding false positives.

Can you add `engine.tabs.label2` to the list?

flodolo:
Actually, hold on. I want to check the code first, because I feel like the logic should be able to catch these better.

flodolo:
I should really learn to read twice before commenting :-\

## decoder.noCodecs.*
We have `.accesskey`, `.button` and `.label`. Since we don't search for `.button`, it stops at `.label` and needs to be manually mapped to a different key.

## pref-irc.open
We have `.accesskey`, `.desc`, `.label` and `.title`. This should be fixed in the logic: if we find a `.label`, we should just stop searching.

This is replaced by #931. Note that it won't show `engine.tabs.`: that's not something that should go in known_mappings, it's something that doesn't get caught by the logic, and it's expected. The reality is that this string should have been called `engine.tabs2.label` :-\

flodolo:
If you have a working local install of Transvision, it would be great if you could test that PR.

flodolo:
> Note that it won't show engine.tabs.: that's not something that should go in known_mappings, it's something that doesn't get caught by the logic, and it's expected. The reality is that this string should have been called engine.tabs2.label :-\

And we actually need to add it, because it falls back to .message now…

PR #936, opened 2018-02-08
Improve FTL support (display, variables)

### LINKED ISSUE #935: Wrong link build to Pontoon
For [this query](https://transvision.mozfr.org/?sourcelocale=en-US&locale=bg&repo=gecko_strings&search_type=strings&recherche=Mozilla+Thunderbird+and+the+Thunderbird+logos%0A+++++++++++++++++++++++++++++++++are+trademarks+of+the+Mozilla+Foundation.&entire_string=entire_string) one of the results have [edit in Pontoon](https://pontoon.mozilla.org/bg/firefox/other-licenses/branding/thunderbird/brand.dtd?search=trademarkInfo.part1) link (to other-license folder) that Pontoon then [redirects](https://pontoon.mozilla.org/bg/firefox/browser/branding/official/brand.dtd/?search=trademarkInfo.part1&string=74131) (to firefox folder). Either hide strings with such wrong links, or build a better one if the string is available in Pontoon.

flodolo:
* Add support for Fluent 0.6 syntax (python-fluent 0.6.2, compare-locales 2.7.0)
* Display expressions as `whitespace: pre-wrap`
schermata 2018-02-08 alle 15 55 31
* Support more variable formats:
    * `{ $foo }`: external argument (already supported)
    * `{ foo }`: message reference
    * `{ -foo }`: term reference

Also fixes #935.

### DIFF
--- a/app/classes/Transvision/AnalyseStrings.php
+++ b/app/classes/Transvision/AnalyseStrings.php
@@ -47,7 +47,7 @@ class AnalyseStrings
 
         $patterns = [
             'dtd'        => '/&([A-Za-z0-9\.]+);/',                        // &foobar;
-            'ftl'        => '/\{\s*(\$[A-Za-z0-9_]+)\s*\}/u',              // { $foobar } Used in FTL files
+            'ftl'        => '/(?<!\{)\{\s*([\$|-]?[A-Za-z0-9_-]+)\s*\}/u',  // { $foo }, { foo }, { -foo } Used in FTL files
             'ios'        => '/(%(?:[0-9]+\$){0,1}@)/i',                    // %@, but also %1$@, %2$@, etc.
             'l10njs'     => '/\{\{\s*([A-Za-z0-9_]+)\s*\}\}/u',            // {{foobar2}} Used in Loop and PDFViewer
             'printf'     => '/(%(?:[0-9]+\$){0,1}(?:[0-9].){0,1}([sS]))/', // %1$S or %S. %1$0.S and %0.S are valid too
--- a/app/classes/Transvision/ShowResults.php
+++ b/app/classes/Transvision/ShowResults.php
@@ -428,6 +428,13 @@ class ShowResults
                 $transliterate_string_id = 'transliterate_' . $string_id;
             }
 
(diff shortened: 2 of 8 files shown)


flodolo:
@TheoChevalier 
We might have to merge this by the end of the week, depending on when the new strings merge to m-c

I tested the changes in the Entities view, and I can't spot anything breaking. `Utils::secureText` transforms new lines in spaces, so it's not helpful here.

It might be a little bit hard to test, since compare-locales still doesn't have a RELEASE_2_7_0 tag (it should have before EOD).

TheoChevalier:
I can give it a spin tonight/tomorrow, I guess I should wait for compare-locales?
I will look into those `secureText` changes, because I’m always afraid we break some edge cases when touching these functions. Pretty sure it’s just me being paranoid, but you know…

flodolo:
> I will look into those secureText changes, because I’m always afraid we break some edge cases when touching these functions. Pretty sure it’s just me being paranoid, but you know…

I looked into #917, which introduce them, and they still work. ShowResults isn't using them.

At this point, the merge date for the string is next week, so there's more time. compare-locales should be tagged tomorrow.

In case you want to test, you can run setup, then just make sure to run `hg pull -u -r default` in libraries/compares-locales to update to latest master. That would be enough for testing.

flodolo:
2.7.0 is now tagged for compare-locales

flodolo:
> Do we need to care about {-brand-name} in tests, or it’s already covered by {{n}}?

It's [already covered](https://github.com/mozfr/transvision/pull/936/files#diff-c783ed2a89074f2386ac34fe10580f25R176). The updated FTL check also excludes double curly apostrophes, since it's already covered elsewhere.

TheoChevalier:
>It's already covered. The updated FTL check also excludes double curly apostrophes, since it's already covered elsewhere.

I’ve seen that, I was mentioning specifically the version without spaces

flodolo:
Ah, missed that. We could add a test for the future (for all 3 cases), but those files will be likely serialized by Pontoon, so they'll always have a space.

PR #937, opened 2018-02-09
Add more tests for FTL variable checks

flodolo:
And try to fix Travis for PHP 5.6


flodolo:
Looks like Travis doesn't fail all the time now (1 out of 2)

PR #943, opened 2018-03-14
 Strip {ok} tags from mozillaorg, fixes #942

### LINKED ISSUE #942: {ok} tags included in mozilla.org edit links
The Pontoon link here is broken https://transvision.mozfr.org/?recherche=%3Ca+href%3D%22%25(mailto_inclusion)s%22%3Einclusion%40mozilla.com%3C%2Fa%3E&repo=mozilla_org&sourcelocale=en-US&locale=fr&search_type=strings&entire_string=entire_string

It turns out we’re stripping the {ok} tag from the displayed string with `htmlspecialchars()` alone (wut?) https://github.com/mozfr/transvision/blob/master/app/views/results_entities.php#L32

I have to check, but don’t we have a dedicated function somewhere to remove those?

TheoChevalier:


app/scripts/tmx/tmx_mozillaorg
```diff
@@ -44,6 +44,13 @@ foreach (Files::getFoldersInPath(GIT . 'mozilla_org/') as $locale) {
 
         foreach ($strings as $english => $translation) {
             $output = function ($str1, $str2) use ($file, $english, $translation) {
+                if (Strings::endsWith(strtolower($str2), '{ok}')) {
```
flodolo: Since trailing and leading whitespaces are non significant in LANG files, can we just strip it and trim?

```PHP
if (Strings::endsWith(strtolower($str2), '{ok}')) {
    $str2 = trim(str_ireplace('{ok}', '', $str2));
}
```

flodolo: Or even simpler

```php
if (Strings::endsWith(strtolower($str2), '{ok}')) {
    $str2 = trim(rtrim($str2, '{ok}'));
}
```

TheoChevalier: Hm, without stripping the trailing space, this query doesn’t return the same result (no perfect match for the locale) Do we care?
http://localhost:8082/?recherche=%3Ca+href%3D%22%25(mailto_inclusion)s%22%3Einclusion%40mozilla.com%3C%2Fa%3E&repo=mozilla_org&sourcelocale=en-US&locale=fr&search_type=strings&entire_string=entire_string

TheoChevalier: oh, just noticed the trim, nevermind

flodolo:
Interesting, we also return one more result at this point, which makes absolutely sense.

PR #945, opened 2018-03-22
Add Python 3 compatibility, switch to virtualenv for dependencies

flodolo:
I think this is ready for testing. The patches to `tmx_products.py` matches this [pull request](https://github.com/flodolo/tmx_maker/pull/11) (with tests).


flodolo:
## Testing instructions

### Python 2

* Update master, run `setup.sh` and `glossaire.sh no-snapshot force-tmx`.
* Rename the `TMX` folder to `TMX_master`

* Remove `compare-locales` and `python-fluent` from the `libraries` folder, apply this patch.
* Run `setup.sh` (it will create a virtualenv, likely with python 2, in `python-venv`) and `glossaire.sh no-snapshot no-update`. The `no-update` ensures that sources don’t change between runs. `force-tmx` is not needed is the `TMX` folder is empty.
* Compare master’s and new `TMX` folders: `diff -r TMX TMX_master` (it should be empty).
* Rename the `TMX` folder as `TMX_venv2`

### Python 3

For Python 3, you need at least Python 3.5, since compare-locales is not available for 3.4. On Debian 8 I used [Miniconda](https://conda.io/docs/user-guide/install/index.html), and created the virtualenv with `virtualenv --python=/home/flodolo/miniconda3/bin/python3.6 python-venv`.

* Remove the `python-venv` folder. Manually create a virtualenv with python3 in the same folder with `virtualenv -p python3 python-venv`.
* Run `setup.sh` (it will use the new virtualenv, and install dependencies) and `glossaire.sh no-snapshot no-update`.

Here things become more interesting. The diff is big, so I created a repo to see the differences, and trying to figure out if it's my fault, or compare-locales.
https://github.com/flodolo/test_python3_transvision/commit/d63c8c01bdbcfe1fe610a35a1b9b4e7adea207fe

flodolo:
cc @Pike. The good new is that there's no PHP around, just Bash and bad Python :)

From most of the diffs:
* Trailing white spaces are trimmed.
* `es-AR` is more interesting. There are differences in new lines: the Python 2 file has `CR+LF`, the new one seem to only have CR. The [actual file](https://hg.mozilla.org/l10n-central/es-AR/file/default/browser/chrome/overrides/netError.dtd#l189) has `CR`, so Python 3 seems to do the right thing? Or maybe we shouldn't have CR in localized files?
*`it` is interesting too (`defines.inc`). Both strings have a `CR` at the end.

I haven't look throughtfully across all locales, but I think it boils down to trimmed whitespaces, and `CR` characters.

flodolo:
Latest change is untested, but tests pass in the other repo: looking at compare-locales code I realized I could use `six.text_type()` to replace `unicode()`

flodolo:
I need to test again:
* I've added [tests for whitespaces](url), and they pass on all versions of Python (2.7, 3.5, 3.6)
* [Additional tests](https://github.com/flodolo/tmx_maker/pull/11/commits/fb5bc80ee02fdc212fb0e62c6582521456b54238) confirm the change in behavior with `CR`

flodolo:
I'm making progress:
* Tests pass on the VM with the same version of Python 3.6, so it's not some weird problem with that version.
* Python 2 and 3 [behave differently](https://github.com/flodolo/tmx_maker/pull/11/commits/62c95afbd6066c91a26e64417f20ad256e42e378) for the Bulgarian file (one of the diffs)

No clue where to go from here.

flodolo:
> Python 2 and 3 behave differently for the Bulgarian file (one of the diffs)

Those are no simple whitespaces (20), they're non-breaking spaces (C2A0)

flodolo:
> Those are not simple whitespaces (20), they're non-breaking spaces (C2A0)

FIled https://bugzilla.mozilla.org/show_bug.cgi?id=1448256

flodolo:
Forgot to add that I tested the latest version with Python 2, and results are identical to current master.

app/scripts/clean_data.py
```diff
@@ -6,8 +6,23 @@
...
+try:
+    dict.iteritems
+except AttributeError:
+    # Python 3
+    def iteritems(d):
+        return iter(d.items())
+else:
+    # Python 2
+    def iteritems(d):
+        return d.iteritems()
```
Pike: `six` has helpers for both of these. Guess you added six after this?

Also further down in the patch.

flodolo: This one is on purpose. 

Unlike the main script (`tmx_products.py`), this one is called directly from bash, as stand-alone, and I don't want to force having `six` as dependency.

flodolo:
We need to wait for that bug to be fixed. It will likely change the output of Python 2 too (for new lines).

flodolo:
compare-locales 3.1.0 is out, but still has a bug with reading files with broken encoding. Worth waiting for the next dot release.

flodolo:
We have 3.1.1

Doing one final round of testing, but I think it's almost ready to be merged.

app/scripts/clean_data.py
flodolo: On second thoughts, I think it makes sense to make a bash script, assume the virtual env is available, and run this script within it.

PR #946, opened 2018-04-01
Cache logic in complex views (accesskeys, variables, shortcuts)

flodolo:


flodolo:
The diff is much larger than it actually is, because the existing code in the model is indented. 

The idea for all views is quite simple:
1. Cache the results. I'm not using the usual `if (! cache::getKey(xxx))` because an empty array is valid, and would never be read from cache.
2. If cache is present, use the existing code. The only change is: store all the data needed in the view within the array, instead of reusing `$target` and `$source` later.
3. Use `array_column` to get the strings IDs from the results array.
4. Update the views to use only the results array.

Non cached views take from 20 to 30MB, while a cached view often takes less than 1MB.

Also truncate the accesskey view for locales like Hindi, with 1700 results.

Issue #949, opened 2018-07-06
Access Key check looks at <entity>.message instead of <entity>.button


TonnesM:
When using the access keys check on Transvision for my locale (nl), I noticed decoder.decodeError.**accesskey** apparently looks at decoder.decodeWarning.**message** instead of decoder.decodeError.**button**, resulting in an incorrect entry.

I understand this needs to be added as an exception?

flodolo:
I'll wait a few more days to push to prod, in case something else comes up.

PR #950, opened 2018-07-06
(fixes issue #949) Check for .button before .message in accesskeys views

flodolo:
Can't spot any regression in the few locales I checked.


flodolo:
Review ping? 🦄

PR #952, opened 2018-10-12
Update compare-locales to 4.0

flodolo:


flodolo:
Merging, since it's just Python code, and verified locally that there are no differences in the TMX output.

PR #953, opened 2018-10-12
.lang parser: don't store missing strings in TMX (fixes #951)

### LINKED ISSUE #951: Edit in Pontoon does not work if the string is not translated in id-less projects
[This](https://transvision.mozfr.org/?recherche=windows+xp&repo=mozilla_org&sourcelocale=en-US&locale=zam&search_type=strings) is an example search. The translation says "Warning: Empty string" and the [Edit in Pontoon](https://pontoon.mozilla.org/zam/mozillaorg/download_button.lang/?string=7865) link points to the right file, but not the right string, because the search parameter is empty and the string number is not right either.

flodolo:
I didn't change the [DotLang class](https://github.com/mozfr/transvision/blob/master/app/classes/Transvision/Dotlang.php#L112), since we copied it from langchecker.


flodolo:
@TheoChevalier review ping?

PR #998, opened 2021-01-06
Remove testing for PHP 5.6, add PHP 7.0

flodolo:
Production server on mozfr.org is currently running 7.0

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-versions: ['5.6', '7.2', '7.4']
+        php-versions: ['7.0', '7.2', '7.4']
     steps:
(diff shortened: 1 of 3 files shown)


flodolo:
Looks like functional tests are broken on PHP 7.0, but I don't think there's a lot of value in trying to figure out why and fix it.

Also tried quickly PHP 8.0, which requires atoum 4.0. Unfortunately, that's compatible with PHP 7.2 and above, and seems to use different class names.

flodolo:
Checked PHP 8.0 in more details. We currently have `use atoum`, and then each class extends `atoum/test`.

In order to work with atoum 4.0, we need `use atoum/atoum`. So, in order to test 8.0, we need to move to atoum 4.0 and drop support for 7.0.

Since [7.0 is unsupported anyway](https://www.php.net/supported-versions.php), I'm tempted to say that's worth doing it anyway, even if the prod server is still on 7.0

PR #999, opened 2021-01-06
Enable PHP 8.0 and 7.3 in testing, drop 7.0

flodolo:


.github/workflows/tests.yaml
```diff
@@ -52,15 +52,13 @@ jobs:
         run: eslint web/js
 
       - name: Run PHP linter (php-cs-fixer)
-        run: vendor/bin/php-cs-fixer --diff --dry-run -v fix
+        run: PHP_CS_FIXER_IGNORE_ENV=1 vendor/bin/php-cs-fixer --diff --dry-run -v fix
```
flodolo, line 55: Apparently PHP 8.0 is still not officially supported
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5371/files

PR #1032, opened 2023-04-06
Commit composer.lock

enderandpeter:
# The Problem

This project is not committing the `composer.lock`. [Composer **strongly recommends** committing this file](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control), if not insists or commands one to do so.

In my web dev career, I often encounter confusion among colleagues about how to handle this file, but in the end Composer's recommendations are essential to follow. I do not believe there is confusion here. I think this project is encouraging environments to make strong assumptions that can easily result in instability.

The `composer.lock` is not too unlike pip's `requirements.txt`. So, why commit this file but not `composer.lock`? We should apply the same reasoning here so we know what exact versions of dependencies will support this app, rather than taking unnecessary risks.

[mozilla-l10n-query](https://github.com/mozilla-l10n/mozilla-l10n-query) is riding the lightning, in my humble opinion. However, I will admit that in my experience, I have often seen Composer packages remain rather stable within major versions among a small set of reliable vendors.

Still, for the list of dependencies in [this project's composer.json](https://github.com/mozfr/transvision/blob/main/composer.json), even for Mozilla-maintained repos, those are too many assumptions to make that updated versions will not cause problems.

# The Solution

* Commit the `composer.lock`. Ideally, this should be the one from the production server, just to lower the risk that any Composer dependency updates have introduced undiscovered unwanted behavior.
* Remove `composer.lock` from `.gitignore` so it is committed.
* Update `CONTRIBUTING.md` to refer to installing locked dependency versions.

Thank you for considering this. I would love to discuss this more.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ cache/lastdataupdate.txt
 cache/stats_*.json
 cache/tag.txt
 cache/version.txt
-composer.lock
 composer.phar
 logs/*.log
 logs/github_log.txt
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,12 +51,12 @@ git checkout -b my_new_branch
 ```bash
 start.sh -tests
 ```
-- Update dependencies with composer:
+- Install dependencies with composer:
 ```bash
-php composer.phar update
+php composer.phar install
 ```
 or, if Composer is installed globally:
 ```bash
-composer update
+composer install
 ```
 - Check our [Coding Standards](https://github.com/mozfr/transvision/wiki/Code-conventions) before submitting pull requests.
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,2612 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
+        "This file is @generated automatically"
+    ],
+    "content-hash": "542ad397807c60b70558116eb0c71edb",
+    "packages": [
+        {
+            "name": "gettext/gettext",
+            "version": "v5.7.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/php-gettext/Gettext.git",
+                "reference": "8657e580747bb3baacccdcebe69cac094661e404"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/php-gettext/Gettext/zipball/8657e580747bb3baacccdcebe69cac094661e404",
+                "reference": "8657e580747bb3baacccdcebe69cac094661e404",
+                "shasum": ""
+            },
+            "require": {
+                "gettext/languages": "^2.3",
+                "php": "^7.2|^8.0"
+            },
+            "require-dev": {
+                "brick/varexporter": "^0.3.5",
+                "friendsofphp/php-cs-fixer": "^3.2",
+                "oscarotero/php-cs-fixer-config": "^2.0",
+                "phpunit/phpunit": "^8.0|^9.0",
+                "squizlabs/php_codesniffer": "^3.0"
+            },
+            "type": "library",
+            "autoload": {
+                "psr-4": {
+                    "Gettext\\": "src"
+                }
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "MIT"
+            ],
+            "authors": [
+                {
+                    "name": "Oscar Otero",
+                    "email": "oom@oscarotero.com",
+                    "homepage": "http://oscarotero.com",
+                    "role": "Developer"
+                }
+            ],
+            "description": "PHP gettext manager",
+            "homepage": "https://github.com/php-gettext/Gettext",
+            "keywords": [
+                "JS",
+                "gettext",
+                "i18n",
+                "mo",
+                "po",
+                "translation"
+            ],
+            "support": {
+                "email": "oom@oscarotero.com",
+                "issues": "https://github.com/php-gettext/Gettext/issues",
+                "source": "https://github.com/php-gettext/Gettext/tree/v5.7.0"
+            },
+            "funding": [
+                {
+                    "url": "https://paypal.me/oscarotero",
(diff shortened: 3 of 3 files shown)


flodolo:
> [mozilla-l10n-query](https://github.com/mozilla-l10n/mozilla-l10n-query) is riding the lightning, in my humble opinion.

Not familiar with the expression… What do you mean? Besides php-cs-fixed, which has no impact on run-time, that project [doesn't use any external library](https://github.com/mozilla-l10n/mozilla-l10n-query/blob/main/composer.json).

enderandpeter:
@flodolo I guess I am referring more to the general practice I see at [setup.sh#L121](https://github.com/mozfr/transvision/blob/main/app/scripts/setup.sh#L121), where an update is done if the `composer.lock` is not present. I do see how with [mozilla-l10n-query](https://github.com/mozilla-l10n/mozilla-l10n-query), the only non dev dependency is very internally maintained, so I suppose you are right that it's not too much like riding lightning (doing something very risky) in the case of such a very focused web service.

Although risk is mitigated in that case, it would be best to lock the dependencies as a common practice. Things might remain stable with this project's required dependencies for a time, but I've seen sudden updates that throw so many other things off rather suddenly.

flodolo:
You can see that tests are failing, because the versions in composer.lock are not suitable for 7.4 (and 8.0 from a quick look). Is there a way to avoid that? At least 8.0 is not [EOLed yet](https://www.php.net/supported-versions.php)

flodolo:
One other question: there's a [GitHub hook running](https://github.com/mozfr/transvision/blob/main/web/github_hook.php#L79-L85) once the repository gets updated. 

Would that result in `composer.lock` changing locally? I assume it wouldn't, as it would be only install new packages based on that file.

enderandpeter:
These are excellent points to raise, indeed. I think I see the reasoning for the lack of composer.lock that is worth discussing more.

This project has all the appearances and behavior of a web application that will be deployed to a server to provide all its translation services. In such a situation, there would only be tests for the version of PHP that this web application is deployed on.

I think I am seeing how this project is also being treated like [mozillal10n/json](https://github.com/mozilla-l10n/phpclass-json), [mozillal10n/cache](https://github.com/mozilla-l10n/phpclass-cache) etc., which more understandably lack a `composer.lock` because they are meant to be used as Composer packages themselves. Such packages that are meant to be used in a wide variety of projects understandably have such tests that check a range of PHP versions. This is a very different expectation than a web service with a finite set of known and maintained deployments.

And so, even though I think that treating this app like such a package seems like a bit too big of scope, I was wanting to know your thoughts before proceeding. If Transvision is meant to be an app that can be deployed on PHP 7.4 and up on any server, then I would actually not recommend a `composer.lock`, because this app is not being used in the same manner as Composer's recommendation on tracking that file. Perhaps Transvision is somewhere in between.

By the way, running `composer update` almost always changes `composer.lock`, if not just its content-hash. The typical expectation is that when devs add/update composer packages, and then test their changes of course, they will run either `composer require [package]` or `composer update [package]` which makes changes to the composer.lock. Then other environments simply run `composer install`, which will read the checked-in `composer.lock`, to get those changes into the `vendor` folder in the environment.

This is what would be done in a typical standalone web application. But I am getting the impression that Transvision is meant to be provided like a web app package, you might say.

flodolo:
>  If Transvision is meant to be an app that can be deployed on PHP 7.4 and up on any server, then I would actually not recommend a composer.lock, because this app is not being used in the same manner as Composer's recommendation on tracking that file. Perhaps Transvision is somewhere in between.

We should probably drop 7.4 at this point, but we can't pin it to a single version of PHP.

Transvision is deployed on different servers managed by different folks (on top of devs potentially having different PHP versions). There are at least 3 public instances on 2 servers at the moment ([my own](https://transvision.flod.org/), and the [official mozfr](https://transvision.mozfr.org/)).

In general, the expectation is to enforce a minimum supported version, so that we can take advantage of specific features. And we typically add support early-ish for new versions, in case one of these servers.

enderandpeter:
@flodolo Thank you for explaining the deployment and the reasoning.

If anything, yes I agree with not supporting PHP 7.4 anymore. Things are just moving too fast. However, from what I am understanding regarding how this app is used, there are good reasons to pin this to a single PHP version.

The intention is for this app to be on the specific deployments you mentioned, especially the three public instances, to support Mozilla's goals as opposed to this app being a tool that is meant to support multiple clients that vary in PHP versions. If clients/users are using different PHP versions, that is is one thing. 

But devs should *not* be using different PHP versions for a web app that is meant to be deployed in well-known managed environments. Everyone should coordinate to understand what is required to make the app run reliably. All deployment environments should likewise be kept up-to-date.

When you say a minimum version is enforced to take advantage of specific features, surely those features are still available in later PHP? If they are deprecated features, then surely for good reason, and there would have to be alternatives.

I should note that the `requirements.txt` is in a slightly different situation, because this app is expecting `python3`, and pip's lock file is not as strict as Composer's.

I do want to continue this conversation, but it might be best if I close the ticket for now because an essential prerequisite would be establishing a common PHP version for the many deployments, for which the latest version would be ideal. I'd love to revisit this, especially after I get Docker image going.

flodolo:
> But devs should _not_ be using different PHP versions for a web app that is meant to be deployed in well-known managed environments.

As an open source project, forcing devs to have a specific PHP version can create an entry barrier.

Also note that this is not an official Mozilla project running on Mozilla systems, it's a community based project running on servers maintained by volunteers (and running other things). Again, can't really enforce a specific version, we just can make sure there is test coverage.

enderandpeter:
I wouldn't call such requirements an entry barrier. They are simply the requirements for the application. It's a bad idea to deploy such an application to whatever server you happen to have lying around, instead of specifically provisioning a server to support an application. VMs and containers are the way to go. When you use them, then you are not limited to what happens to be on your host machine. You can provision web servers to have whatever versions of whatever software is required. I assure you, that is the great benefit of Docker and why I've been on the container bandwagon for so long. Once you get on the Docker bandwagon and see that you can have total control, then you realize just how much freedom you have to define a web app for a specific context.

Issue #827, opened 2016-11-16
@@missing@@ in bug reports


TheoChevalier:
Just noticed we’re sending @@missing@@ to Bugzilla when filing a bug from a missing string, we should probably display something else.

Can be tested on https://transvision.mozfr.org/?sourcelocale=en-US&locale=fr&repo=aurora&search_type=entities&recherche=mail/chrome/messenger/addressbook/abContactsPanel.dtd:mailingListPropertiesMenu.label&perfect_match=perfect_match

rajivjhaimin:
@TheoChevalier  I would like to work on the issue.

flodolo:
Hi, feel free to start and ask questions if you have any.

flodolo:
Updated link: https://transvision.mozfr.org/?recherche=firstrun_urlbar_message2&repo=gecko_strings&sourcelocale=en-US&locale=de&search_type=entities

ayushpratap:
I would like to take it up , can someone tell me how to reproduce the bug and what should be the expected output?

flodolo:
> how to reproduce the bug and what should be the expected output

Open https://transvision.mozfr.org/?recherche=securitySiteDataCookiesOnly&repo=gecko_strings&sourcelocale=en-US&locale=de&search_type=entities and click on <report a bug>

The open bug report shouldn't have `@@ missing @@`

```
The string:
Yes, cookies

Is translated as:
@@missing@@
```

Either write `(missing translation)` or drop that paragraph, the former would be much easier.

flodolo:
> I checked the above-mentioned link and

I'm not following. Did you click the link to report a bug, under "Warning: Missing string"?

ayushpratap:
@flodolo I wanted to work on this bug. So I asked how to reproduce it.
Then you gave the following link
[THE LINK]( https://transvision.mozfr.org/?recherche=securitySiteDataCookiesOnly&repo=gecko_strings&sourcelocale=en-US&locale=de&search_type=entities)

Whereas per you, the string " Yes, cookies " should be " @@missing@@" but instead in place of "@@missing@@" it is "Warning: Missing string"

flodolo:
This is the initial comment

> Just noticed we’re sending @@missing@@ to Bugzilla when filing a bug from a missing string, we should probably display something else.

As expained:
[...]

PR #1033, opened 2023-04-07
Use different Bugzilla comment when translation is missing

enderandpeter:
#827 

# The Problem

After clicking the "report a bug" link for a missing target translation, the Bugzilla comment contains the **@@missing@@** tag used internally by this application.

# The Solution

When the Bugzilla comment is created for the "report a bug" link, and the target translation is missing, use a default comment that pertains to this situation, for example:

```
Source string:
Content crashed

This string has not been translated yet. Proposed translation:
```

# Notes

I have a simple solution in place that will use the above template if the `$target_string` in `Bugzilla::reportErrorLink` is `@@missing@@`. `tests/units/Transvision/Bugzilla.php` has been updated to test for this case. See if you can think of other cases we may want to test.

This is my first contribution to this project, so I apologize if I missed something. I did run `php-cs-fixer` on the changes files. I hope the other tests still pass, but I will do what is necessary to fix any problems. Let me know what you think. Thank you!


app/classes/Transvision/Bugzilla.php
```diff
@@ -55,14 +55,26 @@ public static function reportErrorLink($locale, $entity, $source_string, $target
...
+
+        $default_message = "The string:\n{$source_string}\n\n"
+            . "Is translated as:\n{$target_string}\n\n"
+            . "And should be:\n\n\n\n"
+            . "Feedback via Transvision:\n"
+            . $transvision_url;
+
+        $message = $default_message;
+
+        // If the target string has the @@missing@@ tag, use an alternate comment
```
flodolo, line 67:
```suggestion
        // If the target string has the @@missing@@ tag, use an alternative comment
```

flodolo (changes requested):
This looks great and works, but it also made me realize there is more space for improvement. In the almost 10 years since this code was originally written, Bugzilla has introduced support for Markdown, which can help us a lot here.

We can use code blocks for source/translation, and an actual link instead of the huge Transvision URL. Also, to simplify code a bit, we can use the same preamble (`Source string:`) for both scenarios.

Example:

```php
$message = "Source string:\n\n```\n{$source_string}\n```\n\n";
if($target_string !== '@@missing@@') {
    $message .= "Is translated as:\n\n```\n{$target_string}\n```\n\n"
        . "And should be:\n\n```\n(add your translation here)\n```\n\n";
} else {
    // If the target string has the @@missing@@ tag, use an alternative comment
    $message .= "This string has not been translated yet. Proposed translation:\n\n"
        . "```\n(add your translation here)\n```\n\n";
}
// Add link to the string in Transvision
$message .= "Feedback via [Transvision]({$transvision_url}).";
$bug_message = rawurlencode(
    html_entity_decode($message)
);
```

Feel free to change/improve it, but that should be the general idea (only tested with a couple of strings).

enderandpeter:
I was able to incorporate the changes you suggested. Things appear to be working as expected. Here are some previews of the Bugzilla comment:

## New comment, string missing

New Bugzilla Comment_String Missing

## String Not Missing

New Bugzilla Comment_String Not Missing

## Complex String Missing

New Bugzilla Comment_Complex String Missing

## Complex String Not Missing

New Bugzilla Comment_Complex String Not Missing

I'm wondering if the newlines insides the preformatted blocks should be removed? It appears they are creating several spaces instead of an actual newline.

flodolo:
> Actually, I do see the newlines now. I think the indentation threw me off. Not sure where that is from.

Not sure if [this](https://transvision.flod.org/?sourcelocale=en-US&locale=af&repo=all_projects&search_type=entities&recherche=android_l10n/mozilla-mobile/android-components/components/browser/errorpages/src/main/res/values/strings.xml:mozac_browser_errorpages_safe_browsing_malware_uri_message&entire_string=entire_string) is the string in your second example, but there are spaces in the source

```
    <string name="mozac_browser_errorpages_safe_browsing_malware_uri_message"><![CDATA[
      <p>The site at %1$s has been reported as an attack site and has been blocked based on your security preferences.</p>
    ]]></string>
```

If you inspect the results table, and set `white-space: pre`, you'll see them.

Screenshot 2023-04-10 alle 07 35 14

flodolo (approved):
Congrats for your first contribution.

This will be available in a few minutes in my own server (running beta), and deployed in production in the coming days (hopefully we can figure out the issue with All Projects and deploy both).

enderandpeter:
Heck yeah. Thank you so much!

